PR #6, opened 2015-09-16
Payments list

richardaanzee:
Added a function to the base to get a list of payments where you can set the count and offset, respecting previous functionality.


richardaanzee:
I also fixed issue #5 about being able to refund.

lvgunst:
Todo:
* Add `Payment.canBeRefunded`: Check for attribute `amountRefunded` existence. https://www.mollie.com/en/docs/payments#payment-object
* Add `Payment.canBePartiallyRefunded`: `Payment.canBeRefunded() && Payment.method != 'creditcard'`
* Bail out of `Payments.refund` when payment does not support refunds or partial refunds.
* Add refund example

PR #12, opened 2016-04-01
Properly handling request errors

Pagebakers:
Mollie makes my app crash when requests to Mollie fail because request errors aren't handled.

I added an error handler so this doesn't happen anymore, also rewritten the code to use convention style callbacks `callback(error, result)`

Another solution would be to return the request object in `Client::callRest`.

What do you think?


RicardoJ:
Hello Eelco,

Thank you for your contribution to our node api client. Your purposed pull request seems to be the proper solution to better handle errors within the API client.

Thank you again for your contribution and I'll release a new version tomorrow.

Ricardo,
Mollie

Pagebakers:
Awesome, thanks for releasing this :)

RicardoJ:
#13 After careful review of the merged changes I'll have to revert this in order to review this more carefully with my colleagues. It seem a good contribution, however, just by passing a null as first parameter to ignore the errors will result in breaking those who are using using this api client.

Pagebakers:
The change would require a new major version. It's more inline with nodejs conventions, so I would be in favour. Another benefit is we can use the error parameter not only for request errors, but also to catch errors from api responses.

lvgunst:
@Pagebakers You are right, a new major version is required for this change. Before we release a new version I think we also need to use this error parameter for API error responses like https://github.com/mollie/mollie-api-node/blob/master/src%2Flib%2Fmollie%2Fapi%2Fresource%2Fbase.coffee#L84 Im not sure if this error parameter should always be an instance of `Error`. Do you know?

And while we're at it. Maybe we should also add support for query-string parameters like `count` and `offset`. See #6

Pagebakers:
I'm not sure what the format of the error response is, but you could do something like this
```js
return callback(new Error(body.error.code, body.error.message));
```

Issue #1, opened 2014-05-21
A HTTP error will cause Node.js  to exit (non-gracefully)


swijnands:
If a HTTP error occurs (for whatever reason) during a Mollie API Request, Node.js exits, without a way for the end-user to catch the error, i.e.:

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: Parse Error
    at CleartextStream.socketOnData (http.js:1583:20)
    at CleartextStream.read [as _read] (tls.js:507:12)
    at CleartextStream.Readable.read (_stream_readable.js:320:10)
    at EncryptedStream.write [as _write] (tls.js:366:25)
    at doWrite (_stream_writable.js:221:10)
    at writeOrBuffer (_stream_writable.js:211:5)
    at EncryptedStream.Writable.write (_stream_writable.js:180:11)
    at write (_stream_readable.js:583:24)
    at flow (_stream_readable.js:592:7)
    at Socket.pipeOnReadable (_stream_readable.js:624:5)
```

This can/should be prevented by adding a request.on('error', function(err) { .... }) handler, and calling the end-user callback function with the appropriate feedback, in the rest client:

https://github.com/mollie/mollie-api-node/blob/master/src/lib/mollie/api/client.coffee#L82-92

arenddeboer:
Is this issue still present ? I am bit concerned about using the Mollie Node API in production while this bug exists.

RicardoJ:
Hello @arenddeboer, this issue is still present, however, @Pagebakers created a pull request on April 1th, 2016 (#12) where this is been fixed. Take a look in the purposed pull request for more information.

dabroek:
The fix in #12 was merged, but later reverted in https://github.com/mollie/mollie-api-node/commit/8b520f5ac19878f4a649f626cfc821a45fc7d2eb. This issue still needs to be addressed and resolved.

dabroek:
#18 is addressing this issue.

vernondegoede:
Please use our new `@mollie/api-client` module, which is completely rewritten to ES6 and fully tested.

You can find information on how to upgrade in our [Migration guide](https://github.com/mollie/mollie-api-node/blob/master/MIGRATION.md).

PR #18, opened 2016-10-07
Adressing some stability issues

Hless:
I needed issue #1 fixed for a project I'm  working, so I adressed this while trying to maintain compatibility with current installations. At the moment any uncaught HTTP error causes node to exit non gracefully. The proposed change:
- Return an object with the error key set to the resulting HTTP error. instead of crashing. Any callback checking the error within object should handle this correctly.

Also:
The delete endpoint on the mandates API returns an empty response body, also causing the application to crash. Adressed by checking the body length and parsing the JSON string within a Try/catch statement (the latter is extraneous, perhaps).

### DIFF
--- a/src/lib/mollie/api/client.coffee
+++ b/src/lib/mollie/api/client.coffee
@@ -93,3 +93,5 @@ module.exports = class Client
 
 		request.write(JSON.stringify(data));
 		request.end();
+
+		return request;
--- a/src/lib/mollie/api/client.coffee
+++ b/src/lib/mollie/api/client.coffee
@@ -87,7 +87,17 @@ module.exports = class Client
 				body += data.toString();
 			);
 			response.on("end", ->
-				callback(JSON.parse(body))
+        if body.length
+          parsedBody = {}
+          try
+              parsedBody = JSON.parse(body)
+          catch e
+              console.log(e)
+
+          callback(parsedBody)
+        else
+          callback({})
+
 			);
 		);
 
--- a/src/lib/mollie/api/resource/base.coffee
+++ b/src/lib/mollie/api/resource/base.coffee
@@ -51,7 +51,9 @@ module.exports = class Base
 				if (body.error)
 					return callback(body);
 				callback(this.copy(body, new this.constructor.object));
-		);
+		).on("error", (error) => 
+            callback(error: error)
+        );
 
 	get: (id, callback) ->
 		this.api.callRest(
@@ -59,7 +61,9 @@ module.exports = class Base
 				if (body.error)
 					return callback(body);
 				callback(this.copy(body, new this.constructor.object));
-		);
+		).on("error", (error) => 
+            callback(error: error)
+        );
 
 	update: (id, data, callback) ->
 		this.api.callRest(
@@ -67,7 +71,9 @@ module.exports = class Base
 				if (body.error)
 					return callback(body);
 				callback(this.copy(body, new this.constructor.object));
-		);
+		).on("error", (error) => 
+            callback(error: error)
+        );
 
 	delete: (id, callback) ->
 		this.api.callRest(
@@ -75,7 +81,9 @@ module.exports = class Base
 				if (body.error)
 					return callback(body);
 				callback(this.copy(body, new this.constructor.object));
-		);
+		).on("error", (error) => 
+            callback(error: error)
+        );
 
 	all: (callback) ->
 		this.api.callRest(
@@ -92,7 +100,9 @@ module.exports = class Base
 					list.push(this.copy(body.data[item], new this.constructor.object));
 
 				callback(list);
-		);
+		).on("error", (error) => 
+            callback(error: error)
+        );
 
 	getResourceName: () ->
 		if (this.constructor.resource.indexOf("_") >= 0)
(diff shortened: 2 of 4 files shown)


willemstuursma:
Thanks @Hless. LGTM. I'll ask our node expert @Thijs-Riezebeek to have a look at it Monday, I think we can merge it without issues. Can you hold out that long?

Hless:
My applications are puling the node module from my fork, so in I'm no rush. 

Just thought I'd create a PR for others to benefit

araker:
Any update on when this pull request will be merged?

src/lib/mollie/api/resource/base.coffee
```diff
@@ -51,31 +51,39 @@ module.exports = class Base
 				if (body.error)
 					return callback(body);
 				callback(this.copy(body, new this.constructor.object));
-		);
+		).on("error", (error) =>
```
dabroek: If you do nothing else but returning the error object. There is no need to provide the request object to the resource base class. You can simply call the callback with the error object in the client. Right now you are just introducing a lot of duplicated code in this file.

src/lib/mollie/api/client.coffee
```diff
@@ -87,9 +87,21 @@ module.exports = class Client
 				body += data.toString();
 			);
 			response.on("end", ->
-				callback(JSON.parse(body))
+        if body.length
+          parsedBody = {}
+          try
+              parsedBody = JSON.parse(body)
+          catch e
+              console.log(e)
```
dabroek: A side-effect like a `console.log` is not wanted here. This error needs to be handled correctly. The `try-catch` around the `JSON.parse` is a good idea though.

src/lib/mollie/api/client.coffee
```diff
@@ -87,9 +87,21 @@ module.exports = class Client
...
+        else
+          callback({})
+
 			);
 		);
 
 		request.write(JSON.stringify(data));
 		request.end();
+
+		return request;
```
dabroek: This one is not needed. You can simply add: 
```coffeescript
request.on("error", (error) => callback(error: error)
```

src/lib/mollie/api/client.coffee
```diff
@@ -87,9 +87,21 @@ module.exports = class Client
...
+        if body.length
+          parsedBody = {}
+          try
+              parsedBody = JSON.parse(body)
+          catch e
+              console.log(e)
+
+          callback(parsedBody)
+        else
+          callback({})
```
dabroek: An empty object is not what is returned from the API. I don't think it is a good idea to fake it.

src/lib/mollie/api/resource/base.coffee
Hless: Valid point. However the callback parameter of the methods (get, put, post, etc.) in base class are not exposed to the callrest method in the client. Any suggestion how to work around this? Perhaps, provide an additional argument to callrest with an error callback funciton?

src/lib/mollie/api/client.coffee
Hless: See the first issue raised in your review. Probably need to pass in an errorcallback argument.

src/lib/mollie/api/client.coffee
Hless: Any other suggestion? Just return the error to the user?

src/lib/mollie/api/client.coffee
Hless: This may be true, but I believe the client is copying the models defaults (empty keys) on to the object that the API returns. So the resulting object would still be a valid model.

Check this line in the base class:
```callback(this.copy(body, new this.constructor.object));```

Comment (Hless) [inline src/lib/mollie/api/client.coffee]: So I wrote it like this in order to maintain backwards compatibility, but perhaps I should add an error key into the object?

**Comment by Hless** [comment]
Thanks for the review. Little bit swamped at the moment, but I will address the issues somewhere next week.

File: src/lib/mollie/api/client.coffee (hunk 2 above)
Comment (dabroek) [inline src/lib/mollie/api/client.coffee]: Return the error in the `callback`.

File: src/lib/mollie/api/client.coffee (hunk 4 above)
Comment (dabroek) [inline src/lib/mollie/api/client.coffee]: You are right, this is our bad design. Your solution might be the simplest fix right now.

File: src/lib/mollie/api/client.coffee (hunk 3 above)
Comment (dabroek) [inline src/lib/mollie/api/client.coffee]: In this API the error is set inside the error property of the callback. We are currently rewriting the API where we will support promises and will implement the standard `(error, result) => style` callback.

File: src/lib/mollie/api/resource/base.coffee (hunk 1 above)
Comment (dabroek) [inline src/lib/mollie/api/resource/base.coffee]: It is. When you return a callback with an error parameter, the methods (create, get, update, delete, and all) check if it is set and return it in the callback. 
```coffeescript
if (body.error)
  return callback(body);
```

TMeerhof:
Today our production server had multiple non-graceful exits because off the missing try catch block on the JSON.parse(body). I know we can do more to check the body before we send it, but it would be a nice addition. Is there any progress on this pull request?

dabroek:
@TMeerhof There is a new version available written in ES6 that does not suffer from this problem. It also allows you to use promises, among other things. Just `npm install mollie-api-node@next`.

PR #19, opened 2016-10-07
Fixes #11; no way to pass count to methods.all

### LINKED ISSUE #11: No way to pass count to methods.all
When calling Mollie.methods.all, you will always get a maximum of 10 results back, without any way to retrieve more. There is no way to pass a count parameter.

```
Base.prototype.all = function(callback) {
      return this.api.callRest("GET", this.constructor.resource, null, null, (function(_this) {
```
^ the data parameter is hardcoded to null

Would be nice to implement this

(See https://www.mollie.com/en/docs/reference/methods/list)

Kenanjl:


willemstuursma:
LGTM.

Thijs-Riezebeek:
I've merged this change locally and fixed the feedback I left above.

PR #26, opened 2017-06-08
support for count and offset in all function

maxx-dev:
In the "all" function the querystring parameters (count, offset) are currently not supported.
Example: https://www.mollie.com/en/docs/reference/customers/list


dabroek:
You've been editing the `lib` folder? Why don't you try to edit the source and run the build script?

PR #28, opened 2017-06-12
Updated mandate object, fixes #27

### LINKED ISSUE #27: mandate.isValid() always returns false
https://github.com/mollie/mollie-api-node/blob/master/src/lib/mollie/api/object/mandate.coffee#L45 compiles to:

```js
(function() {
  var Mandate;

  module.exports = Mandate = (function() {
    Mandate.STATUS_VALID = "valid";

    Mandate.STATUS_INVALID = "invalid";

    function Mandate() {
      this.resource = "mandate";
      this.id = null;
      this.status = null;
      this.method = null;
      this.customerId = null;
      this.details = null;
      this.createdDatetime = null;
    }

    Mandate.prototype.isValid = function() {
      return this.status === this.STATUS_VALID;
    };

    return Mandate;

  })();

}).call(this);
```

I believe the `return this.status === this.STATUS_VALID;` line should be `return this.status === Mandate.STATUS_VALID;` (use class variable instead of instance variable).

jookes:
This commit fixes the issue discussed in #27. npm run build has been used to compile .coffee into .js. 

I tested it on my integration and it returns true when the mandate is valid


jookes:
@dabroek Would you be able to pull this into the repository? This way I can use the regular version again 😄

PR #40, opened 2017-11-02
Implement Chargeback resource

vernondegoede:
- Support `https://api.mollie.nl/v1/chargebacks` endpoint (https://www.mollie.com/nl/docs/reference/chargebacks/list-all)
- Added unit tests
- Added integration tests

### DIFF
--- a/tests/integration/chargebacks.test.js
+++ b/tests/integration/chargebacks.test.js
@@ -4,6 +4,10 @@ import dotenv from 'dotenv';
 
 import mollie from '../../dist/mollie';
 
+console.log = function(msg) {
+  process.stdout.write(`${msg}\n`);
+};
+
 /**
  * Overwrite the default XMLHttpRequestAdapter
  */
--- a/tests/integration/chargebacks.test.js
+++ b/tests/integration/chargebacks.test.js
@@ -1,0 +1,31 @@
+import axios from 'axios';
+import httpAdapter from 'axios/lib/adapters/http';
+import dotenv from 'dotenv';
+
+import mollie from '../../dist/mollie';
+
+/**
+ * Overwrite the default XMLHttpRequestAdapter
+ */
+axios.defaults.adapter = httpAdapter;
+
+/**
+ * Load the API_KEY environment variable
+ */
+dotenv.config();
+
+const mollieClient = mollie({ apiKey: process.env.API_KEY });
+
+describe('customers', () => {
+  it('should integrate', done =>
+    mollieClient.chargebacks.all()
+      .then((chargebacks) => {
+        expect(chargebacks).toBeDefined();
+        done();
+      })
+      .catch((err) => {
+        expect(err).toBeNull();
+        done();
+      }),
+  );
+});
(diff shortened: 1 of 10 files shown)


tests/unit/models/chargeback.test.js
```diff
@@ -0,0 +1,23 @@
+import Chargeback from 'models/chargeback';
+
+describe('customer model', () => {
```
lvgunst: Fix model name

vernondegoede: Oops, renamed. ✅

tests/integration/chargebacks.test.js
```diff
@@ -0,0 +1,31 @@
...
+axios.defaults.adapter = httpAdapter;
+
+/**
+ * Load the API_KEY environment variable
+ */
+dotenv.config();
+
+const mollieClient = mollie({ apiKey: process.env.API_KEY });
+
+describe('customers', () => {
```
dabroek: Customers?

vernondegoede:
I'm not sure why the Travis build is failing. It's working fine locally using a `.env` file. Travis should be adding the `API_KEY` as a secret env variable, but for some reason this doesn't work. 

I asked @wvdeutekom to look into this.

tests/integration/chargebacks.test.js
vernondegoede: Fixed.

dabroek:
I don't think this should be a top level resource. Chargeback is a sub-resource. You should implement these:
```
v1/payments/<payment_id>/chargebacks
v1/payments/<payment_id>/chargebacks/<chargeback_id>
```

vernondegoede:
@dabroek I'll add the chargeback subresources in another PR, as discussed.

tests/integration/chargebacks.test.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+axios.defaults.adapter = httpAdapter;
+
+/**
+ * Load the API_KEY environment variable
+ */
+dotenv.config();
+
+const mollieClient = mollie({ apiKey: process.env.API_KEY });
+
```
wvdeutekom: @vernondegoede in order to start debugging this you will have to print two things here:
* All environment variables this application can read, this way we know if `API_KEY` is set at all in the pipeline (even though the settings say it is)
* The full mollieClient object, so we can see if the apiKey variable is set if the env variable does exist

lvgunst: I don't think we ever set an `API_KEY` environment variable in Travis config (using a encrypted variable): https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

@dabroek Have you ever get this to work on the `mollie:next` branch?

dabroek: Yes I had it working before. I put the environment variable there myself.

wvdeutekom: I could see the `API_KEY` in the settings, except for the value assigned to it.

RickWong:
LGTM! 🎉

PR #57, opened 2018-06-01
Update the cacert.pem

Smitsel:


Smitsel:
Hi @dabroek,

Can you have a look for me and see to it that this change is released? :)

willemstuursma:
This file is used to validate the authenticity of the certificate that is provided by Mollie when connecting.

Keeping this file up to date ensures no MITM attacks can happen.

vernondegoede:
The master branch will be released as 2.0.0 soon (currently on NPM as `2.0.0-alpha.1
`). This change should be made [against the v1.3.5 tag](https://github.com/mollie/mollie-api-node/tree/v1.3.5) as well and be released as a patch.

PR #60, opened 2018-06-15
Add Mollie ESlint config and Prettier.

vernondegoede:


lvgunst:
Have you run the prettier / eslint formatter on the code? No changes?

src/models/method.js
```diff
@@ -1,35 +1,35 @@
...
+  static KBC = "kbc";
+  static GIFTCARD = "giftcard";
+  static INGHOMEPAY = "inghomepay";
 
   constructor(props) {
     super(props);
 
     const defaults = {
-      resource: 'method',
+      resource: "method",
```
lvgunst: I see a lot of changes to use double quotes while you have configured single quotes in `.prettierrc`. We prefer single quotes in all our projects.

vernondegoede: Yeah, there's something wrong. I ran Prettier again this morning (just to be sure) and it doesn't seem to recognize the `.prettierrc` file. Will check it again later today 👍

vernondegoede:
@lvgunst Can you take a look again? Ran both Prettier + ESLint. Config seems to work now. I changed the package name to `@mollie/eslint-config-node`.

src/models/customer.js
```diff
@@ -8,7 +8,7 @@ export default class Customer extends Model {
     super(props);
 
     const defaults = {
-      resource: 'customer',
+      resource: "customer",
```
lvgunst: Somehow it's still using double quotes.

src/models/customer.js
```diff
@@ -20,8 +20,8 @@ export default class Customer extends Model {
       _links: {
         mandates: null,
         subscriptions: null,
-        payments: null,
-      },
+        payments: null
```
lvgunst: Also the `"trailingComma": "all"` is not working

PR #83, opened 2018-09-07
allow count & offset parameters in all() method

karsveling:
this is a bugfix in order to let count & offset to work.

in src/resource.js

--> we need to load payments in batches but there was a bug that offset + count was not passed to the api. This is what we used at LessonUp to be able to iterate.

### DIFF
--- a/src/resource.js
+++ b/src/resource.js
@@ -164,9 +164,15 @@ export default class Resource {
     if (typeof params === 'function') {
       cb = params; // eslint-disable-line no-param-reassign
     }
-
+    
+    // offset and count parameters should be passed as queryParameters, so they have a special treatment.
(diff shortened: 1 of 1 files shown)


vernondegoede:
Hi @karsveling,

Thanks for your interest in the Node client! I think you're referring to the Mollie v1 API. This package was completely rewritten a couple of months ago and works with the v2 API.

The v2 API works using cursor pagination, while the v1 API uses `count` and `offset` to paginate. For more information, please check out our [Pagination guide](https://docs.mollie.com/guides/pagination).

If you'd like to use the v1 API, you can use the v1 API client: https://github.com/mollie/mollie-api-node/tree/v1.4.0.

Please let us know if you have any further questions.

PR #84, opened 2018-09-07
Update resource.js

karsveling:
this is a bugfix in order to let count & offset to work.

in src/resource.js

--> we need to load payments in batches but there was a bug that offset + count was not passed to the api. This is what we used at LessonUp to be able to iterate.


vernondegoede:
Duplicate of #83.

PR #85, opened 2018-09-19
Convert amount from number to string in README example

Reinoptland:
Hi! I'm trying out this package and I spotted a small mistake in the docs:

## What this pullrequest does:

- Corrects a small mistake in the README example. The amount is a number instead of a string which will make the API respond with a 422.


vernondegoede:
Thanks @Reinoptland 👍

PR #86, opened 2018-09-20
Change webhook call from 'GET' to 'POST'

tania-dm:
Hi,

I was testing the API and I came across a small mistake in the way the webhook is called in the example file. Shouldn't the method be 'POST'? If you try with 'GET' you get a 404.

Thanks,
Tania


vernondegoede:
You're right! Thanks @tania-dm 👍

dabroek:
Hi @tania-dm !

First of all, thank you for your interest in our node client!

You are absolutely correct, we use a POST request when posting to your webhook, as [described in our documentation](https://docs.mollie.com/guides/webhooks). Thank you so much for contributing to this package by opening a PR. Any help is greatly appreciated, not only by us, but also by the rest of the community.

Hope to see more of your contributions in the near future!

PR #93, opened 2018-10-09
Automatic deployments using Travis

vernondegoede:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mollie/api-client",
-  "version": "2.1.0",
+  "version": "2.1.0-alpha",
```
martijnpolder: alpha seems a versions that would be released as a test version prior to the v2.1.0?

vernondegoede: What version number would you suggest? alpha releases won't be installed by NPM automatically, so we can safely remove the tag from Git and this [specific version from NPM afterwards](https://docs.npmjs.com/cli/unpublish).

vernondegoede: I'm not sure why Travis doesn't run the tests though.

vernondegoede:
@martijnpolder I'm discussed the testing plan with @dabroek and we'll see if this works when we actually publish a new version.

.travis.yml
```diff
@@ -1,9 +1,18 @@
...
+cache: yarn
 script:
    - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then yarn run test:unit:cov; fi'
    - 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then yarn run test:cov; fi'
+deploy:
+  provider: npm
+  api_key:
+    secure: "fpsIqEZNvzcS2g91vwuR8UJo+4Ok6xGuVV/uL5nNia+8cnA+cSfQEmPm4w+0gB3F1UIOGs9ItuMqcBlZOMe7oZxr9Wi+J05LfZGL4cibRXU+ydpnQ0d5Si7/p4W3vskA+SnolDjHPCHAVxZiyRUXswuhMnajbrAejeBvS2n0WF4="
+  email: "systems@mollie.com"
+  skip_cleanup: true
```
vernondegoede: We still want the files to be available during the "deploy" step.

.travis.yml
```diff
@@ -1,9 +1,18 @@
...
   - "4.8"
+cache: yarn
 script:
    - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then yarn run test:unit:cov; fi'
    - 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then yarn run test:cov; fi'
+deploy:
+  provider: npm
+  api_key:
+    secure: "fpsIqEZNvzcS2g91vwuR8UJo+4Ok6xGuVV/uL5nNia+8cnA+cSfQEmPm4w+0gB3F1UIOGs9ItuMqcBlZOMe7oZxr9Wi+J05LfZGL4cibRXU+ydpnQ0d5Si7/p4W3vskA+SnolDjHPCHAVxZiyRUXswuhMnajbrAejeBvS2n0WF4="
+  email: "systems@mollie.com"
```
vernondegoede: Check this with @martijnpolder. This email address is already publicly available through the NPM registry, so it's fine to use it in `.travis.yml`

Issue #95, opened 2018-10-11
Implement missing endpoints


vernondegoede:
- [ ] https://docs.mollie.com/reference/v2/chargebacks-api/get-chargeback

PR #96, opened 2018-10-11
Improve examples

vernondegoede:
Based on the way we structure our examples in the Ruby client.

While I was creating the examples, I noticed that we are missing the sub-resource to list chargebacks for a payment. I created an issue to keep track of this: https://github.com/mollie/mollie-api-node/issues/95.

The "integration examples" are now moved to a "with-express" sub-folder, which makes more sense in my opinion.

When this PR is approved, I'll update our api-documentation and add the examples over there as well.


examples/methods/list.js
```diff
@@ -0,0 +1,15 @@
+/**
```
lvgunst: Can you also add examples for:
- https://docs.mollie.com/orders/migrate-to-orders#retrieving-payment-methods
- https://docs.mollie.com/payments/multicurrency#filtering-payment-methods

vernondegoede: This client doesn't support Orders (yet). I'll add the filtering example though 👍

vernondegoede:
@lvgunst Added the example 👍

PR #97, opened 2018-11-03
API Update: Orders API, fix QR Codes, Rollup update, Typescript and more

firstred:
Originally this was supposed to consist of two simple updates to introduce TypeScript declarations and the Orders API, so that this client can be used in the (upcoming) Reaction Commerce plugin, but somehow, somewhere, this escalated really quickly!

This "simple update" has turned into a huge PR for the node API client, including but not limited to:
- [x] TypeScript
  - [x] Add TypeScript declarations
  - [x] Migrate Rollup
  - [x] Migrate Jest
  - [x] Migrate ESLint to TSLint (~not all alternative rules have been found, yet~ we don't need them)
  - [x] Define input parameters (e.g. payments.create), so we can immediately see in our IDEs what is expected 
- [x] Orders API
  - [x] Resources
  - [x] TypeScript declarations
  - [x] Models
  - [x] Unit tests
  - [x] Integration tests
- [x] Jest
  - [x] TypeScript checks
  - [x] No implicit any ( define all the things :smile: )
- [x] Examples
  - [x] Remain in plain JavaScript
  - [x] Orders API
- [x] Browser support (some frameworks, such as Reaction Commerce, based on Meteor.js,  automatically split code between the server and client)
  - [x] Decouple cacert.pem
  - [x] No more node.js only references
- [x] Rollup
  - [x] Export CommonJS
  - [x] Export ES6 modules
  - [x] Export TypeScript declarations 
 - [x] Fix QR Codes
  - [x] Restore `include.details=qrCode` query

Please don't merge the changed at this moment. If there's anything that needs to be changed or left out, it can still be updated from my own "typescript only" branch. Also, the tests do not cover every endpoint, yet. As always, the boring part comes last :slightly_smiling_face:


vernondegoede:
This is amazing! Thanks 💯 
Please let us know when you're done, so we start reviewing!

mattijsbliek:
Wow this is great work 🙌

I would really like to have separate PRs for each of the points listed in the title though. This will make reviewing much more manageable, and we can already start merging the smaller stuff in. I get that this is a big ask, so if you need someone from Mollie to work with you on this let us know 🙂

martijnpolder:
@vernondegoede I confirmed the `cacert.pem` file in commit 863c497 to be untampered with.

firstred:
Hi all,
Thanks for your enthusiasm!

I'll try and see if I can split this huge PR into multiple ones. This one was originally made to quickly apply the changes to the Reaction Commerce module, but I can imagine what a mess that must be to review ^^

If interested I can share the Reaction Commerce module, too, so you'll have a real life example.

Note that this is a premature pull request. I usually request changes too early, just to figure out if my proposals will be accepted. Are there any items that would have to be changed or added before creating the new PRs? Is it okay for the whole codebase to become TypeScript? TIA for any answers!

vernondegoede:
> Is it okay for the whole codebase to become TypeScript?

Yes, definitely. 👍 TypeScript will prevent bugs and makes it easier for others to use our API client.

> If interested I can share the Reaction Commerce module, too, so you'll have a real life example.

Sure, feel free to add us to your repo!

JFGHT:
Great job in here, looking forward to it.

vernondegoede:
@firstred Is there anything I can help you with that would enable you to finish up the last tasks in this PR? Let me know, I'd be happy to help!

bertlobbezoo:
I'm looking forward to the Orders API implementation.

firstred:
Sorry for the late response. I'm afraid that I do not have enough time to split this pull request. I encountered several problems during my attempts, because every item so highly interconnected with the others.

I would propose to just finish this pull request as there are only a few more items left. Reviewing might take a bit longer, but in the end it looks like we'll be saving a lot of time by keeping everything one pull request.

**Short status update**
I did manage to use this PR (Payments API only) in Reaction Commerce and didn't encounter any issues (repo coming soon!).
So far so good, but the tests for the Orders API need to be completed as well to be sure that part works, too.

@epdenouden I noticed an edit in the issue description. The item `Restore include.details=qrCode query` has been unchecked, but I can't find an explanation. Did this happen by accident?

vernondegoede:
@firstred No worries. Feel free to finish off the last couple of open tasks and let us know when you're ready for a review. 👍

firstred:
Will do! Thanks!

firstred:
Nice, it looks like we're getting somewhere! I had some difficulties with publishing the TypeScript type definitions in an NPM package, something I've never done before, but it looks like it is coming along nicely after the last refactor:

*PhpStorm/WebStorm/IntelliJ*
molliephpstorm

*Visual Studio Code*
mollievscode

Notice how deep IntelliSense is now working and how it even triggers boths IDEs to show a documentation popup.

The Orders API is still a work in progress. I managed to connect a few endpoints, but there are still a few that do not work. Luckily TypeScript seems to be pointing out all the flaws at the moment, so there's our first TypeScript reward already :slightly_smiling_face: 

The tests have been migrated to TypeScript as well to make it easier to quickly write tests for the missing endpoints. When that is finished we should be good to go!

package.json
```diff
@@ -43,38 +43,36 @@
...
-    "rollup-plugin-url": "^2.0.1",
-    "ts-jest": "^23.10.4",
-    "ts-node": "^7.0.1",
+    "rollup-plugin-node-resolve": "^4.0.0",
+    "rollup-plugin-typescript2": "^0.19.2",
+    "ts-jest": "^23.10.5",
+    "ts-node": "^8.0.1",
     "tslib": "^1.9.3",
-    "tslint": "^5.11.0",
+    "tslint": "^5.12.1",
```
vernondegoede: Can we remove `tslint` and start using the [TypeScript ESLint plugin](https://github.com/typescript-eslint/typescript-eslint)?

https://eslint.org/blog/2019/01/future-typescript-eslint

firstred: Just came across the exact same article. It looks like the better option indeed! Let's migrate to ESLint then.

firstred: Turned out to be a great idea! I have updated the [eslint repo](https://github.com/mollie/eslint-config-mollie).

firstred:
Ready to review!

src/models/Order.ts
```diff
@@ -0,0 +1,67 @@
...
+  public canceledAt?: string;
+  public completedAt?: string;
+  public amountCaptured?: IAmount | null;
+  public amountRefunded?: IAmount | null;
+  public consumerDateOfBirth?: string;
+
+  /**
+   * Order constructor
+   *
+   * @public ✓ This method is part of the public API
```
vernondegoede:
```suggestion
   * @public ✓ This constructor is part of the public API
```

src/models/Order.ts
```diff
@@ -0,0 +1,67 @@
+import Model from '../model';
+import { IOrder } from '../types/order';
+import { IAmount } from '../types/global';
+import Payment from './Payment';
+
+/**
+ * The `order` model
```
vernondegoede:
```suggestion
 * The `Order` model
```

src/models/OrderLine.ts
```diff
@@ -0,0 +1,54 @@
...
+  public shippableQuantity = null;
+  public refundableQuantity = null;
+  public cancelableQuantity = null;
+  public createdAt = null;
+  public metadata = null;
+
+  /**
+   * OrderLine constructor
+   *
+   * @public ✓ This method is part of the public API
```
vernondegoede:
```suggestion
   * @public ✓ This constructor is part of the public API
```

src/models/OrderLine.ts
```diff
@@ -0,0 +1,54 @@
+import Model from '../model';
+import { IOrderLine } from '../types/order/line';
+
+/**
+ * The `orderline` model
```
vernondegoede:
```suggestion
 * The `OrderLine` model
```

src/models/Payment.ts
```diff
@@ -0,0 +1,162 @@
...
+    customer: null;
+  };
+  public _embedded = null;
+  public amountCaptured = null;
+  public orderId = null;
+
+  /**
+   * Payment constructor
+   *
+   * @public ✓ This method is part of the public API
```
vernondegoede:
```suggestion
   * @public ✓ This constructor is part of the public API
```

src/models/Order.ts
```diff
@@ -0,0 +1,67 @@
...
+   * Order constructor
+   *
+   * @public ✓ This method is part of the public API
+   */
+  public constructor(props?: Partial<IOrder>) {
+    super();
+
+    Object.assign(this, props);
+
+    if (this._embedded != null && typeof this._embedded === 'object') {
```
vernondegoede: It's also possible to embed `refunds`: https://docs.mollie.com/reference/v2/orders-api/get-order#embedding-of-related-resources

src/models/Shipment.ts
```diff
@@ -0,0 +1,37 @@
+import Model from '../model';
+import { IShipment } from '../types/shipment';
+
+/**
+ * The `shipment` model
```
vernondegoede:
```suggestion
 * The `Shipment` model
```

src/models/Shipment.ts
```diff
@@ -0,0 +1,37 @@
...
+    documentation: null,
+  };
+
+  // Access token parameters
+  public testmode?: boolean;
+
+  /**
+   * Shipment constructor
+   *
+   * @public ✓ This method is part of the public API
```
vernondegoede:
```suggestion
   * @public ✓ This constructor is part of the public API
```

src/models/Subscription.ts
```diff
@@ -0,0 +1,105 @@
...
+  public mandateId = null;
+  public nextPaymentDate = null;
+
+  // Access token parameters
+  public testmode?: boolean;
+
+  /**
+   * Subscription constructor
+   *
+   * @public ✓ This method is part of the public API
```
vernondegoede:
```suggestion
   * @public ✓ This constructor is part of the public API
```

src/models/Shipment.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * The `shipment` model
+ *
+ * {@link IShipment}
+ */
+export default class Shipment extends Model implements IShipment {
+  public static resourcePrefix = 'shp_';
+
+  public resource = null;
```
vernondegoede: The `resource` is already known, therefore it can be `shipment` by default.

src/resource.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+    return error;
+  }
+
+  /**
+   * Set the parent ID by providing the parent
+   *
+   * @since 1.1.1
+   *
+   * @deprecated This method is not available
```
vernondegoede:
```suggestion
   * @deprecated since 2.2.0
```

src/resource.ts
```diff
@@ -0,0 +1,340 @@
...
+   *
+   * @since 1.0.0
+   *
+   * @public ✓ This method is part of the public API
+   */
+  public async create(prms: any, cb?: ResourceCallback): Promise<Model> {
+    const params = cloneDeep(prms);
+    const callback = typeof params === 'function' ? params : cb;
+    const query: any = {};
+    if (params !== null && typeof params === 'object') {
```
vernondegoede: Please use the `isPlainObject` method from `lodash`, as arrays are also objects:

```
typeof []
> "object"
```

src/resource.ts
```diff
@@ -0,0 +1,340 @@
...
+import { AxiosInstance, AxiosResponse } from 'axios';
+import { cloneDeep } from 'lodash';
+
+import Model from './model';
+import List from './models/List';
+
+/**
+ * @deprecated since 2.2.0
+ */
+export type ResourceCallback = (error: any, resource: any) => void;
```
vernondegoede: Thanks, I'm okay with deprecating the callback. Most people use `async / await` nowadays anyway.

src/resource.ts
```diff
@@ -0,0 +1,340 @@
...
+   *
+   * @returns Resource list
+   *
+   * @since 1.0.0
+   */
+  public async list(prms?: any, cb?: ResourceCallback): Promise<List<Model>> {
+    const params = cloneDeep(prms);
+    try {
+      const query: any = {};
+      if (params != null && typeof params === 'object') {
```
vernondegoede: Can you bring this entire class under tests? Especially the embeds / include parts.

.npmignore
```diff
@@ -1,11 +0,0 @@
-*.log
```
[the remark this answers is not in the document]
firstred: Strange, I moved to a whitelist system in the `packageJson.files` array, but I can't find it anymore :O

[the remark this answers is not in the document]
firstred: Restore the npmignore (blacklists) or move to the files array (whitelist)? The latter worked better for me in tests as I kept forgetting files.

src/resource.ts
firstred: That one gets me every time :slightly_smiling_face:

examples/refunds/cancel.js
```diff
@@ -1,17 +0,0 @@
-/**
```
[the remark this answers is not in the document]
firstred: No clue, they should be restored!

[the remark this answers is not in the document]
firstred: Ahh, now I see. They will have to be replaced by the examples in the documentation, which covers both the Orders API and Payments API: https://github.com/mollie/api-documentation/pull/378

src/resource.ts
firstred: Turns out to be since v2 of the API, actually. Does that correspond to v2.0.0 of the client?

firstred: Seems like it. Shall I change it to `@deprecated 2.2.0. This method is not supported by the v2 API.` to make it clearer?

.npmignore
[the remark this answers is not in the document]
firstred: `.npmignore` (blacklist) has been replaced with the `files` array in `package.json` (whitelist). Is that OK?

src/resource.ts
firstred: Oops, forgot to tell about the deprecation, but thanks for letting me know it is okay to add!

examples/refunds/cancel.js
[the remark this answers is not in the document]
firstred: Restored + added order refund examples

src/resource.ts
vernondegoede: Please do!

src/resources/chargebacks.ts
```diff
@@ -0,0 +1,101 @@
...
+   */
+  public async get(id: string, params: any, cb?: Function): Promise<Chargeback> {
+    throw new ApiException(`The method "get" does not exist on the "${this.apiName}"`);
+  }
+
+  /**
+   * @deprecated This method is not available
+   */
+  public async cancel(): Promise<boolean> {
+    throw new ApiException(`The method "cancel" does not exist on the "${this.apiName}"`);
```
vernondegoede: These exceptions are thrown a lot. It might make sense to introduce a `NotImplementedException`, to which we pass the `this.apiName`. We can find the method name by accessing the stack trace: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack

src/resources/orders/shipments.ts
```diff
@@ -0,0 +1,253 @@
...
+import { CreateCallback, GetCallback, ListCallback, UpdateCallback } from '../../types/shipment/callback';
+import Order from '../../models/Order';
+import Resource from '../../resource';
+
+/**
+ * The `order_shipments` resource
+ *
+ * @since 2.2.0
+ */
+export default class OrdersShipmentsResource extends OrdersBaseResource {
```
vernondegoede: What about `OrderShipmentsResource`?

.eslintrc
```diff
@@ -0,0 +1,42 @@
+{
+  "extends": [
+    "plugin:@typescript-eslint/recommended",
+    "@mollie/eslint-config-base"
```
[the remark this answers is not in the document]
vernondegoede: Just released the new Typescript config: `@mollie/eslint-config-typescript`.

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+  chargebacks: ChargebacksResource;
+  payments_captures: PaymentsCapturesResource;
+  customers: CustomersResource;
+  customers_payments: CustomersPaymentsResource;
+  customers_mandates: CustomersMandatesResource;
+  customers_subscriptions: CustomersSubscriptionsResource;
+  orders: OrdersResource;
+  orders_refunds: OrdersRefundsResource;
+  orders_lines: OrdersLinesResource;
+  orders_shipments: OrdersShipmentsResource;
```
vernondegoede:
```suggestion
  order_shipments: OrderShipmentsResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+  payments_chargebacks: PaymentsChargebacksResource;
+  chargebacks: ChargebacksResource;
+  payments_captures: PaymentsCapturesResource;
+  customers: CustomersResource;
+  customers_payments: CustomersPaymentsResource;
+  customers_mandates: CustomersMandatesResource;
+  customers_subscriptions: CustomersSubscriptionsResource;
+  orders: OrdersResource;
+  orders_refunds: OrdersRefundsResource;
+  orders_lines: OrdersLinesResource;
```
vernondegoede:
```suggestion
  order_lines: OrderLinesResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface IMollieApiClient {
+  payments: PaymentsResource;
+  methods: MethodsResource;
+  payments_refunds: PaymentsRefundsResource;
+  refunds: RefundsResource;
+  payments_chargebacks: PaymentsChargebacksResource;
+  chargebacks: ChargebacksResource;
+  payments_captures: PaymentsCapturesResource;
+  customers: CustomersResource;
+  customers_payments: CustomersPaymentsResource;
```
vernondegoede:
```suggestion
  customer_payments: CustomerPaymentsResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export interface IMollieApiClient {
+  payments: PaymentsResource;
+  methods: MethodsResource;
+  payments_refunds: PaymentsRefundsResource;
+  refunds: RefundsResource;
+  payments_chargebacks: PaymentsChargebacksResource;
+  chargebacks: ChargebacksResource;
+  payments_captures: PaymentsCapturesResource;
```
vernondegoede:
```suggestion
  payment_captures: PaymentCapturesResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+interface ICreateMollieApi {
+  httpClient: AxiosInstance;
+}
+
+export interface IMollieApiClient {
+  payments: PaymentsResource;
+  methods: MethodsResource;
+  payments_refunds: PaymentsRefundsResource;
+  refunds: RefundsResource;
+  payments_chargebacks: PaymentsChargebacksResource;
```
vernondegoede:
```suggestion
  payment_chargebacks: PaymentChargebacksResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+import PaymentsCapturesResource from './resources/payments/captures';
+
+interface ICreateMollieApi {
+  httpClient: AxiosInstance;
+}
+
+export interface IMollieApiClient {
+  payments: PaymentsResource;
+  methods: MethodsResource;
+  payments_refunds: PaymentsRefundsResource;
```
vernondegoede:
```suggestion
  payment_refunds: PaymentRefundsResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+  refunds: RefundsResource;
+  payments_chargebacks: PaymentsChargebacksResource;
+  chargebacks: ChargebacksResource;
+  payments_captures: PaymentsCapturesResource;
+  customers: CustomersResource;
+  customers_payments: CustomersPaymentsResource;
+  customers_mandates: CustomersMandatesResource;
+  customers_subscriptions: CustomersSubscriptionsResource;
+  orders: OrdersResource;
+  orders_refunds: OrdersRefundsResource;
```
vernondegoede:
```suggestion
  order_refunds: OrderRefundsResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+  methods: MethodsResource;
+  payments_refunds: PaymentsRefundsResource;
+  refunds: RefundsResource;
+  payments_chargebacks: PaymentsChargebacksResource;
+  chargebacks: ChargebacksResource;
+  payments_captures: PaymentsCapturesResource;
+  customers: CustomersResource;
+  customers_payments: CustomersPaymentsResource;
+  customers_mandates: CustomersMandatesResource;
+  customers_subscriptions: CustomersSubscriptionsResource;
```
vernondegoede:
```suggestion
  customer_subscriptions: CustomerSubscriptionsResource;
```

src/create-mollie-api.ts
```diff
@@ -0,0 +1,79 @@
...
+  payments: PaymentsResource;
+  methods: MethodsResource;
+  payments_refunds: PaymentsRefundsResource;
+  refunds: RefundsResource;
+  payments_chargebacks: PaymentsChargebacksResource;
+  chargebacks: ChargebacksResource;
+  payments_captures: PaymentsCapturesResource;
+  customers: CustomersResource;
+  customers_payments: CustomersPaymentsResource;
+  customers_mandates: CustomersMandatesResource;
```
vernondegoede:
```suggestion
  customer_mandates: CustomerMandatesResource;
```

src/resources/customers/index.ts
```diff
@@ -0,0 +1,218 @@
...
+      if (!startsWith(id, Customer.resourcePrefix)) {
+        throw Resource.errorHandler(
+          { error: { message: 'The customer id is invalid' } },
+          typeof params === 'function' ? params : cb,
+        );
+      }
+
+      return super.get(
+        id,
+        typeof params === 'function' ? null : params,
```
vernondegoede: We can abstract this into a separate function (`getParameters` or something similar) and remove this duplication from other resources as well.

firstred: `useParameters()` ^^

firstred: I think it might be a good idea to start abstracting this when we can remove the deprecated parts. This call to `super.get` will have to be changed with the next major version anyway.

examples/orders/list.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * @docs https://docs.mollie.com/reference/v2/orders-api/list-orders
+ */
+const mollie = require('@mollie/api-client');
+
+const mollieClient = mollie({ apiKey: 'test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM' });
+
+(async () => {
+  try {
+    const mostRecentOrders = await mollieClient.orders.page();
```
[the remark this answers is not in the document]
firstred: I've tried here to make it look more like the other clients and the documentation. This is why there is both a `.page` and `.list` alias. If that is too confusing, then it might be better to remove the aliases.

src/resources/customers/index.ts
vernondegoede: Hmm, not sure to be honest. I would like to prevent introducing breaking changes as long as possible, especially since we already introduced breaking changes a few months ago. I'm really happy that the typescript migration allows us to stay backward compatible.

examples/orders/list.js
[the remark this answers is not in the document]
vernondegoede: I like the aliases! That part is great. But I think we should be consistent with naming in our examples, as this might confuse new devs trying out this library. Since we use ".all" in other examples as well, this might make more sense to use in this example as well.

src/create-mollie-api.ts
mattijsbliek: Same for `PaymentsCapturesResource` above, should be singular `Payment` 🙂

src/create-http-client.ts
```diff
@@ -0,0 +1,41 @@
...
+    ...options.headers,
+    Authorization: `Bearer ${options.apiKey}`,
+    'Accept-Encoding': 'gzip',
+    'Content-Type': 'application/json',
+    'User-Agent': `node.js/${process.version}`,
+    'X-Mollie-User-Agent': `mollie/${version}`,
+  };
+
+  // Setting the root CA certificate will fail in a browser environment
+  if (typeof window !== 'undefined' && typeof fs !== 'undefined' && typeof fs.readFileSync !== 'undefined') {
```
lvgunst: Shouldn't this be `typeof window === 'undefined'` if you are targeting Node?

CHANGELOG.md
```diff
@@ -4,6 +4,18 @@
 
 All notable changes to this project will be documented in this file.
 
+#### v2.2.0 - 2018-01-21
+  - Migrated to TypeScript
+  - Added full support for the Orders API
+  - Add tests for the Orders API and base classes and a few regression tests to guard against removing deprecated functionality too soon
+  - Added Orders API examples
+  - Add browser support (decoupled `cacert.pem`, remove node.js only dependencies)
```
vernondegoede: @mattijsbliek noted that we should not remove the Node.js only dependencies, since it's a Node API client. 

Doing so will allow users to use this in the browser, which means they will expose their API keys publicly. This is not something we want.

.prettierrc
```diff
@@ -1,6 +0,0 @@
-{
```
vernondegoede: Can we bring back this file? And reformat the examples and the code using this Prettier config?

CHANGELOG.md
vernondegoede: Maybe we should even throw an error when `typeof window !== 'undefined'` upon API client initialization. Our technical support department often gets questions because developers try to use the API client in a browser environment. We can prevent support calls by throwing an `UnsupportedEnvironmentException` :)

src/create-http-client.ts
```diff
@@ -0,0 +1,41 @@
...
+  const newOptions = cloneDeep(options);
+  newOptions.baseURL = 'https://api.mollie.com:443/v2/';
+
+  newOptions.headers = {
+    ...options.headers,
+    Authorization: `Bearer ${options.apiKey}`,
+    'Accept-Encoding': 'gzip',
+    'Content-Type': 'application/json',
+    'User-Agent': `node.js/${process.version}`,
+    'X-Mollie-User-Agent': `mollie/${version}`,
```
lvgunst: I was checking our backend to see the metrics for the Node 2.x client but couldn't find any. 

I found the problem. We are only parsing the `User-Agent` header, `X-Mollie-User-Agent` is ignored. We currently also do not recognize `node.js` or `mollie`, only `Node` and `Mollie`.

Can you please change the user agent header to `Node/${process.version} Mollie/${version}`?

.prettierrc
firstred: Oops, I thought it was lacking TypeScript support. Looks like all we have to do is upgrade prettier.

CHANGELOG.md
firstred: I'll double check what parts are added to the browser build by Meteor.js. I had to remove a couple of node.js dependencies to make it work on Reaction Commerce.
Maybe those dependencies would only have to be removed from the entry file, but the way Meteor handles it all is magic to me right now.

firstred: Meteor.js automatically splits files between browser and server, that is what I can tell from what I've seen in the final production build, but I wonder what files it picks. It might be just the entry file that ends up in the browser build.

src/create-http-client.ts
firstred: You are absolutely right! :open_mouth:

firstred: Thanks!

CHANGELOG.md
vernondegoede: I'm okay with removing these dependencies, if we don't need them.
In that case we should definitely throw an error when the global `window` reference is available, to prevent people from using it within in the browser.

firstred: Ahh, that might have happened in older commonjs only builds. In those builds everything ends up in one large file and there is no way to exclude code in browser builds (tree shaking). Reaction Commerce has tree shaking enabled by default, which means that it should now use `jsnext:main` or `module` as an entry point in `package.json` and be able to prune whatever's not imported in the [`client` folder](https://github.com/mollie/mollie-reaction-commerce/tree/aa1800138958bfb498609ab3017c1635893bed9d/client) of a Reaction Commerce plugin.

firstred: Just a thought! Have yet to confirm.

src/resources/customers/index.ts
firstred: Yeah, we should only add new functionality. I meant postponing these kinds of abstractions until v3 or any other major version. Best to keep it as is for now.

src/resources/customers/subscriptions/index.ts
```diff
@@ -0,0 +1,281 @@
...
+      this.setParentId(customerId);
+
+      return super.list(
+        typeof params === 'function' ? null : params,
+        typeof params === 'function' ? params : cb,
+      ) as Promise<List<Subscription>>;
+    }
+
+    const { customerId, ...parameters } = defaults(params, { customerId: this.parentId });
+    if (!startsWith(customerId, Customer.resourcePrefix)) {
```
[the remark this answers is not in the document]
firstred: lodash's `startsWith` handles more than just strings, so we can can throw a consistent error for other types as well.

examples/orders/list.js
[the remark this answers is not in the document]
firstred: Using `.all` in all examples now.

src/create-http-client.ts
firstred: Sure, I have updated the `User-Agent`.

src/resources/orders/shipments.ts
firstred: I've tried to make these consistent with the actual API endpoint:

```
/**
 * The `customers_mandates` resource
 *
 * @since 1.2.0
 */
export default class CustomersMandatesResource
```
but if that is an issue, we can change the names.

package.json
```diff
@@ -17,79 +17,112 @@
...
   "scripts": {
-    "prepublish": "yarn run build",
-    "precommit": "lint-staged",
+    "prepublish": "yarn build",
     "commit": "git-cz",
     "prepush": "yarn test",
-    "test": "yarn run build && jest",
-    "test:watch": "yarn run build && jest --watch",
-    "test:cov": "yarn run build && jest --coverage",
+    "test": "jest",
```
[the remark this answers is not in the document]
firstred: Unfortunately, I haven't been able to run the new tests on anything belong node `6.9.0`.

It looks like the `ts-jest` transformer does not actually work on lower versions. Also, Babel 7 requires `>=6.9.0` or higher.

This gives us two options:
- Increase the minimum requirements (node `4.8` => node `6.9.0`)
- Downgrade to (much) older tooling and convert the `tests` folder back to node `4.8` JavaScript.

I have increased the minimum requirements for now, so we can test the actual build.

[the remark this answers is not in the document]
firstred: Creating the actual build itself will always require node `>= 9.0.0`. It's best to leverage `nvm` and just switch back and forth between every build and test.

.prettierrc
firstred: I have restored prettier and integrated it with ESLint to prevent configuration conflict. Note that we might have to disable several ESLint style rules in the future. It is now prettier for style rules + ESLint for the remaining rules.

src/resources/chargebacks.ts
firstred: I have introduced a `NotImplementedException`

CHANGELOG.md
firstred: ESM build no longer looks like to be an issue with Reaction Commerce, nice!

firstred:
I have added a `PUBLISH_TO_NPM` environment variable to pick the build that can deploy to npm. Not sure if we might be running multiple builds on the same node version, so figured this would be a nice alternative.

firstred:
`addVersionString()` method added.

rollup.config.js
```diff
@@ -1,16 +1,20 @@
...
-import url from 'rollup-plugin-url';
+import typescript from 'rollup-plugin-typescript2';
+import copy from 'rollup-plugin-copy-glob';
 
 export default {
-  entry: 'src/mollie.js',
-  format: 'cjs',
-  dest: 'dist/mollie.js',
+  input: 'src/mollie.ts',
+  external: ['fs', 'https', 'path', 'axios', 'qs', 'lodash/omit', 'lodash/toPlainObject'],
```
lvgunst: Shouldn't this be `lodash`? We are only importing from `lodash`, never from `lodash/<function>` directly. 

Are we sure we are not importing the whole `lodash` package? Here are some suggestions https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark

README.md
```diff
@@ -34,7 +34,7 @@
 
 ## Prerequisites
 
-Mollie API client requires Node 4.8.x or higher to be installed.
+Mollie API client requires Node 6.9.x or higher to be installed.
```
lvgunst: Isn't changing the required Node version a breaking change?

lvgunst: I've checked our metrics on the used user-agents since January 1st. The lowest version is Node 6.9.1 with one request. So let's change this without a major version bump.

vernondegoede: 🎉

firstred:
> I'm still new to TypeScript but I really like where this is going. Thanks for this PR!
> 
> I've create a test project and used `npm link` to use this package. I was expecting to get type intellisense but this wasn't working. It couldn't find any type declarations. Is it working for you correctly? Maybe I am missing something.
> 
> I was looking into my problem and I found that the location of the types is defined in the `package.json` and was linking to `dist/esm/src/mollie.d.ts` but this file was missing after the build.
> 
> After setting `declaration` to `true` in `tsconfig.json` the declaration files were created. In the build the declaration files are using the modules aliases (`@mollie-types`, `@models` etc.) which can't be resolved after the build is done. Once I replaced all imports to using relative imports everything seemed to be working fine.
> 
> Do you have another easier solution to solve this? Otherwise I can commit my changes to replace all imports to relative imports and to enable `declarations`.

Great find, thanks! I was having the same issues and after setting `declaration` to `true` things improved a bit. I guess we might have to base our configuration on this example to make everything work: https://github.com/a-tarasyuk/rollup-typescript-babel/

It should be able to handle the aliases, I added them to prevent a couple of bugs I caused earlier with relative imports (`../../../../` gets really confusing after a while), so it would really be nice to be able to keep them.

The actual problem here seems to be that we output everything with TypeScript (w/ `declarations === false`) and are bypassing Babel (it should complain about runtimeHelpers not being enabled, but it doesn't). Let's dive deeper into this problem...

src/resource.ts
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    try {
+      const response: AxiosResponse = await this.getClient().get(`${this.getResourceUrl()}/${id}${qs.stringify(query, { addQueryPrefix: true })}`);
+
+      // noinspection JSPotentiallyInvalidConstructorUsage
+      const model = new (this.constructor as any).model(response.data);
+
+      if (callback) {
+        return callback(null, model);
```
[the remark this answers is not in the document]
firstred: Hmmz, that should be double-checked, indeed. I believe my IDE complained about not returning otherwise, because by marking a function as `async` we are at leasting promising to return a `Promise`.

src/types/shipment/params.ts
```diff
@@ -0,0 +1,76 @@
...
+ *                   become required for this endpoint.
+ *
+ * @param testmode - Set this to `true` to make this order a test shipment.
+ *
+ * @see https://docs.mollie.com/reference/v2/shipments-api/create-shipment
+ */
+export interface ICreateParams {
+  orderId: string;
+
+  lines: Array<IOrderLine>;
```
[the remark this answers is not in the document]
firstred: Thanks for clarifying!

rollup.config.js
firstred: Correct, thanks! I'll update it.

JFGHT:
What happened to this?

PR #118, opened 2019-09-03
Update ESLint rules. Ignore several rules and fix others.

vernondegoede:


src/createMollieClient.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/camelcase */
```
Marinolinderhof: I think it's better to make an exception on line level instead of the entire file. Case the rule does make sense right?

src/createMollieClient.ts
```diff
@@ -46,7 +47,8 @@ function createHttpClient(options: MollieOptions): AxiosInstance {
   axiosOptions.headers['Accept-Encoding'] = 'gzip';
   axiosOptions.headers['Content-Type'] = 'application/json';
 
-  var customVersionStrings = options.versionStrings;
+  let customVersionStrings = options.versionStrings;
+
   if (undefined == customVersionStrings) {
```
Marinolinderhof: we use Yoda? and shoudn't this be `===`?

src/createMollieClient.ts
```diff
@@ -46,7 +47,8 @@ function createHttpClient(options: MollieOptions): AxiosInstance {
...
     customVersionStrings = [];
   } else if (false == Array.isArray(customVersionStrings)) {
```
Marinolinderhof: yoda?

src/createMollieClient.ts
```diff
@@ -46,7 +47,8 @@ function createHttpClient(options: MollieOptions): AxiosInstance {
   axiosOptions.headers['Accept-Encoding'] = 'gzip';
   axiosOptions.headers['Content-Type'] = 'application/json';
 
-  var customVersionStrings = options.versionStrings;
+  let customVersionStrings = options.versionStrings;
```
Marinolinderhof: let customVersionStrings = options.versionStrings || [];

will help the checks on line 52

src/createMollieClient.ts
```diff
@@ -55,7 +57,7 @@ function createHttpClient(options: MollieOptions): AxiosInstance {
   axiosOptions.headers['User-Agent'] = [
     `Node/${process.version}`,
     `Mollie/${libraryVersion}`,
-    ...(customVersionStrings as Array<string>).map(versionString => {
+    ...(customVersionStrings as string[]).map(versionString => {
```
Marinolinderhof: (customVersionStrings as string[]) I think should always be an array so to cast it is a bit weird imo.

src/createMollieClient.ts
Marinolinderhof: you may remove all the if and replace it with 

```
if (!Array.isArray(options.versionStrings) ) {
  customVersionStrings= [customVersionStrings];
}
```

src/createMollieClient.ts
```diff
@@ -55,7 +57,7 @@ function createHttpClient(options: MollieOptions): AxiosInstance {
...
       //                platform /version
       const matches = /^([^\/]+)\/([^\/\s]+)$/.exec(versionString);
```
Marinolinderhof: some YODA statements underneath line 62

src/models/List.ts
```diff
@@ -57,15 +57,15 @@ export default class List<T> extends Array {
...
+    list.nextPage = async (): Promise<any> =>
       getResources(
         {
           ...params,
           ...List.getNextPageParams(_links),
         },
         callback,
       );
-    list.previousPage = async () =>
+    list.previousPage = async (): Promise<any> =>
```
Marinolinderhof: you know it's a list right? `Promise<IList>`

src/models/List.ts
```diff
@@ -57,15 +57,15 @@ export default class List<T> extends Array {
     const list = new List();
     list.links = _links;
     list.count = count;
-    list.nextPage = async () =>
+    list.nextPage = async (): Promise<any> =>
```
Marinolinderhof: i think it's void?  `Promise<void>`

src/types/customer/index.ts
```diff
@@ -12,8 +12,8 @@ export interface ICustomer {
   name: string;
   email: string;
   locale: Locale;
-  recentlyUsedMethods: Array<PaymentMethod>;
-  metadata: any;
+  recentlyUsedMethods: PaymentMethod[];
+  metadata: Record<string, string>;
```
Marinolinderhof: is this true? `metadata: Record<string, string>;`  ?

src/createMollieClient.ts
Pimm: This, too, depends on how we want this function to handle edge-cases.

`options.versionStrings || [];` would change the behaviour of an empty string or the number 0 being passed as `versionStrings`.

src/createMollieClient.ts
Pimm: That depends on the behaviour we desire should someone ever pass `null` as `versionStrings`.

`==` means it will be treated the same as not passing any version strings. `===` means it will throw an exception, which one could argue is to be expected.

src/types/customer/index.ts
vernondegoede: Yes, this is correct. It can be a JSON object with any key / value pair.

src/createMollieClient.ts
vernondegoede: I disagree. I like that we currently allow providing a single `versionString`, as well as multiple. I do understand your point that it does feel a bit awkward though.

src/models/List.ts
vernondegoede: It's actually:

```
    list.nextPage = async (): Promise<List<Model>> =>
```

I've updated this in the code.

PR #119, opened 2019-09-05
Two small fixes in the documentation

Pimm:
Is this the correct migration guide link? Please check.


vernondegoede:
Thanks @Pimm!

PR #120, opened 2019-09-23
Add typing support for array of payment methods

PastaJ36:


vernondegoede:
We prefer the `Type[]` notation. We might enforce (and auto-fix) this in the future through ESLint.

Thanks both!

PR #132, opened 2020-03-04
add paypal mandate method

BjoernRave:
I saw, that paypal as mandate method was missing in the typescript types, but it is available according to the docs: https://docs.mollie.com/reference/v2/subscriptions-api/create-subscription


vernondegoede:
Sharp – thanks @BjoernRave!

PR #135, opened 2020-03-17
Accept every parameter when fetching a list of resources

vernondegoede:
One of our merchants noticed that the Klarna methods would never appear when fetching methods using the Node.js API client, while they did show up when making the API request using manually.

The reason they would not show up is because only a few of the parameters that are passed in to `Resource.list` method were actually sent to the API. The Klarna payment methods, for example, only show up when passing the option `resource: "orders"` to the Methods API.

This PR allows passing in all parameters to the API, except for the ones that require some **special 🌈** treatment (`include` & `embed`). We shouldn't care about validation for these parameters in the API client, since that's already done in the API itself.

### DIFF
--- a/src/resource.ts
+++ b/src/resource.ts
@@ -213,24 +213,23 @@ export default class Resource {
   public async list(prms?: any, cb?: ResourceCallback): Promise<List<Model>> {
     const params = cloneDeep(prms);
     try {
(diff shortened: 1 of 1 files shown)


DaanKouters:
When is this PR planned to be merged?

PR #136, opened 2020-03-23
Interfaced data

### LINKED ISSUE #122: v3: Various invalid TypeScript definitions
Since v3 TypeScript definitions have been added but some are invalid:

```
@mollie\api-client\dist\types\src\models\OrderLine.d.ts

Error:(22, 5) TS2416: Property '_links' in type 'OrderLine' is not assignable to the same property in base type 'IOrderLine'.
  Type '{ productUrl: null; imageUrl: null; }' is not assignable to type 'IOrderLineLinks'.
    Types of property 'productUrl' are incompatible.
      Type 'null' is not assignable to type 'IUrl'.
```

```
@mollie\api-client\dist\types\src\models\Payment.d.ts

Error:(12, 5) TS2416: Property 'id' in type 'Payment' is not assignable to the same property in base type 'IPayment'.
  Type 'null' is not assignable to type 'string'.
(Repeated)
```

```
@mollie\api-client\dist\types\src\resources\customers\payments.d.ts

Error:(53, 5) TS2416: Property 'create' in type 'CustomersPaymentsResource' is not assignable to the same property in base type 'CustomersBaseResource'.
  Type '(params: ICreateParams, cb?: CreateCallback | undefined) => Promise<Payment>' is not assignable to type '(prms: any, cb?: ResourceCallback | undefined) => Promise<Model>'.
    Type 'Promise<Payment>' is not assignable to type 'Promise<Model>'.
(Repeated)
```

Many of these issues are from interface implementations that override a specifically typed value (`string`) with `null` or other invalid typecasting.

Pimm:
This PR changes the internal architecture of the library.

I have a huge list of ESLint errors to check. **This is not done**. But I would appreciate if you could take a look and let me know if you have any ideas, thoughts, doubts.

The goals of this PR are to:
- export better type definitions, and to
- make the library easier to maintain.

These are the most important changes:

1. Data objects (for instance, payments) coming from the Mollie server are no longer wrapped in a class.
Previously, a payment coming from the server was typed `IPayment` and then turned into a `Payment` (class) instance in the constructor of said class.
With this PR, there is no more class. The library simply injects helpers into payments coming from the server (such as `isOpen`) and returns the plain objects.
   * This should fix #122.
   * Secondly, because a developer cannot call `new Payment(…)`, it becomes more obvious that they should do `mollieClient.payments.create(…)` instead.
   * Lastly, it reduces code duplication.
2. Resources (such as `mollieClient.payments`) implement their own endpoints, instead of relying on the inheritance chain.
Previously, `PaymentResource` would extend a base class which provides `create`, `get` et cetera. That seems sensible, but in practice every resource overrides a method for every endpoint anyway.
With this PR, `PaymentResource` implements `create` itself.
   * As a result, non-existing endpoints (such as `mollieClient.chargebacks.create`) don't exist in the library (and code completion) at all; instead of throwing a `NotImplementedError`. (In fact, `NotImplementedError` is completely eliminated in this PR.)
3. Parameter types are based on the aforementioned interfaces. The parameters for `create` and `update` are usually similar. We don't have to define them from scratch every time. This reduces code duplication.
4. Callback-style calls are supported by adding a single line instead of having a separate branch.

Instead of this:
```javascript
if (typeof params === 'function' || typeof cb === 'function') {
  … // Callback-style code
}

… // Same code, but promise-style
```
We simply have this:
```javascript
if (renege(this, this.create, ...arguments)) return;
… // Promise-style code
```

I have tried to avoid any breaking changes.

Of course, it is still JavaScript. If developers out there are relying on implementation details (such as calling `new Payment(…)` ‒ even though they shouldn't), this will break their code anyway. But the tests we copied over from the PHP version still work, and so should all of the examples.

On the other hand, this should allow developers to remove those `// @ts-ignore` lines.


Pimm:
(Also fixes #137.)

src/errors/ApiError.ts
```diff
@@ -42,7 +32,7 @@ export default class ApiError extends Error {
    *
    * @public ✓ This method is part of the public API
    */
-  public getField(): string {
+  public getField(): string | undefined {
```
vernondegoede: In our internal codebase we use the types below. I would love to use these in here too.

```
declare type Maybe<T> = T | undefined;
declare type Nullable<T> = T | null;
```

src/resources/ParentedResource.ts
```diff
@@ -0,0 +1,32 @@
...
+  protected defaultParentId: string | undefined;
+
+  /**
+   * Returns the passed parent identifier, or `defaultParentId` as set by `withParent` if the former is `undefined`.
+   */
+  protected getParentId(input: string | undefined): string | undefined {
+    if (input != undefined) {
+      return input;
+    } /* if (input == undefined) */ else {
+      return this.defaultParentId;
```
vernondegoede: At Mollie, we prefer as little `else` statements as possible. We can simply return instead of going through the `else` :-)

vernondegoede: Also, I would actually turn statements like these around. If undefined, return the default parent ID.

src/resources/customers/mandates/CustomersMandatesResource.ts
```diff
@@ -0,0 +1,189 @@
...
+   * @since 3.0.0
+   *
+   * @see https://docs.mollie.com/reference/v2/mandates-api/list-mandates
+   *
+   * @public ✓ This method is part of the public API
+   */
+  public list(parameters: ListParameters): Promise<List<Mandate>>;
+  public list(parameters: ListParameters, callback: Callback<List<Mandate>>): void;
+  public list(parameters: ListParameters) {
+    if (renege(this, this.list, ...arguments)) return;
```
vernondegoede: In our internal codebase, we don't allow if-statements like this. We prefer always writing full if-statements.

src/errors/ApiError.ts
Pimm: Cool.

src/resources/ParentedResource.ts
Pimm: I remember now that we had this discussion before! I'll try and internalise this style.

src/resources/customers/mandates/CustomersMandatesResource.ts
Pimm: Personally, I don't like if-statements without the curly brackets either. However, this is very much boilerplace code. If we don't allow an exception here, this goes from one line to three lines **in every endpoint**.

Another option would be to wrap every endpoint. However, I don't think we could do overloading in that way.

PR #142, opened 2020-05-04
Update endpoints

Pimm:
Adds:
- permissions
- organizations
- profiles
- onboarding
- payments.update

### DIFF
--- a/src/plumbing/checkId.ts
+++ b/src/plumbing/checkId.ts
@@ -1,4 +1,4 @@
-type ResourceKind = 'capture' | 'chargeback' | 'customer' | 'mandate' | 'order' | 'orderline' | 'payment' | 'refund' | 'shipment' | 'subscription';
+type ResourceKind = 'capture' | 'chargeback' | 'customer' | 'mandate' | 'order' | 'orderline' | 'organization' | 'payment' | 'permission' | 'profile' | 'refund' | 'shipment' | 'subscription';
 
 const prefixes = new Map<ResourceKind, string>([
   ['capture', 'cpt_'],
@@ -7,17 +7,31 @@ const prefixes = new Map<ResourceKind, string>([
   ['mandate', 'mdt_'],
   ['order', 'ord_'],
   ['orderline', 'odl_'],
+  ['organization', 'org_'],
   ['payment', 'tr_'],
+  ['profile', 'pfl_'],
   ['refund', 're_'],
   ['shipment', 'shp_'],
   ['subscription', 'sub_'],
 ]);
+/**
+ * Retruns whether the passed permission identifier seems plausible (`true`); or is definitely invalid (`false`).
+ */
+const checkPermissionId = (() => {
+  const pattern = /^[a-z.]+$/;
+  return function checkPermissionId(value: string) {
+    return pattern.test(value);
+  };
+})();
 /**
  * Returns whether the passed identifier seems plausible (`true`); or is definitely invalid (`false`).
  */
 export default function checkId(value: string | undefined, resource: ResourceKind): value is string {
   if (value == undefined) {
     return false;
+  }
+  if (resource == 'permission') {
+    return checkPermissionId(value);
   }
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   return value.startsWith(prefixes.get(resource)!);
--- a/src/plumbing/checkId.ts
+++ b/src/plumbing/checkId.ts
@@ -15,23 +15,16 @@ const prefixes = new Map<ResourceKind, string>([
   ['subscription', 'sub_'],
 ]);
 /**
- * Retruns whether the passed permission identifier seems plausible (`true`); or is definitely invalid (`false`).
- */
-const checkPermissionId = (() => {
-  const pattern = /^[a-z.]+$/;
-  return function checkPermissionId(value: string) {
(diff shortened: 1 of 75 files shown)


jamesjjk:
Would be great to confirm and merge?

src/resources/orders/orderlines/OrderLinesResource.ts
```diff
@@ -87,6 +87,6 @@ export default class OrdersLinesResource extends ParentedResource<OrderData, Ord
       throw new ApiError('The order id is invalid');
     }
     const { orderId: _, ...data } = parameters;
-    return this.network.delete(this.getResourceUrl(orderId)) as Promise<true>;
+    return this.network.delete<true>(this.getResourceUrl(orderId));
```
vernondegoede, line 90: What's the reason the `OrdersResource` passes in an `Order` type, while this one passes in `true`?

src/resources/organizations/OrganizationsResource.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  protected injectPrototypes = injectPrototypes;
+
+  /**
+   * Retrieve an organization by its ID.
+   *
+   * @since 3.2.0
+   *
+   * @see https://docs.mollie.com/reference/v2/permissions-api/get-permission
```
vernondegoede, line 23: This should be the organization link.

src/resources/profiles/ProfilesResource.ts
```diff
@@ -0,0 +1,127 @@
...
+   * @see https://docs.mollie.com/reference/v2/profiles-api/update-profile
+   *
+   * @public ✓ This method is part of the public API
+   */
+  public update(id: string, parameters: UpdateParameters): Promise<Profile>;
+  public update(id: string, parameters: UpdateParameters, callback: Callback<Profile>): void;
+  public update(id: string, parameters: UpdateParameters) {
+    if (renege(this, this.update, ...arguments)) return;
+    if (!checkId(id, 'profile')) {
+      throw new ApiError('The payment id is invalid');
```
vernondegoede, line 104:
```suggestion
      throw new ApiError('The profile id is invalid');
```

src/resources/profiles/ProfilesResource.ts
```diff
@@ -0,0 +1,127 @@
...
+   * @see https://docs.mollie.com/reference/v2/profiles-api/get-profile
+   *
+   * @public ✓ This method is part of the public API
+   */
+  public get(id: string): Promise<Profile>;
+  public get(id: string, callback: Callback<Profile>): void;
+  public get(id: string) {
+    if (renege(this, this.get, ...arguments)) return;
+    if (!checkId(id, 'profile')) {
+      throw new ApiError('The payment id is invalid');
```
vernondegoede, line 51:
```suggestion
      throw new ApiError('The profile id is invalid');
```

src/plumbing/checkId.ts
```diff
@@ -7,18 +7,32 @@ const prefixes = new Map<ResourceKind, string>([
...
+  };
+})();
 /**
  * Returns whether the passed identifier seems plausible (`true`); or is definitely invalid (`false`).
  */
 export default function checkId(value: string | undefined, resource: ResourceKind): value is string {
   if (value == undefined) {
     return false;
   }
+  if (resource == 'permission') {
```
vernondegoede, line 33: Hmm, not sure if this is really necessary. Let's let the API handle this.

src/resources/organizations/OrganizationsResource.ts
Pimm: Thanks!

src/resources/profiles/ProfilesResource.ts
Pimm: Great!

src/resources/profiles/ProfilesResource.ts
Pimm: Yes!

src/plumbing/checkId.ts
Pimm: Yes, I agree that this might be a little excessive. I'll take this out.

src/resources/orders/orderlines/OrderLinesResource.ts
Pimm: Because some endpoints return 200 along with the model, while others return 204.

See [cancel order](https://docs.mollie.com/reference/v2/orders-api/cancel-order#response) vs [cancel order lines](https://docs.mollie.com/reference/v2/orders-api/cancel-order-lines#response).

Previously, the delete-y methods returned `Promise<Model | boolean>` according to the type definition. Now we specify which of those ‒ the model or `true` ‒ is returned in that definition.

vernondegoede:
Could you also add unit tests for these endpoints? I understand that the integration tests are a bit harder, as they require valid access tokens.

src/resources/orders/orderlines/OrderLinesResource.ts
Pimm: We could introduce an enum for this if we feel that is helpful in the future.

vernondegoede:
As discussed:
- Tests for endpoints can be ported from the PHP library
- "Documentation" can be removed for all endpoint (but let's keep the `@see` and `@since`)

Issue #147, opened 2020-05-19
Cancel order endpoint testmode parameter not accepted


jamesjjk:
When calling orders.cancel([id], { testmode: true }). it is not accepted by the api there seems to be an issue with calling the delete request with a parameter, it seems it only accepts a form element.

Pimm:
Hi, thanks for creating this issue.

Are you using Mollie Connect/OAuth?
What is the error that you get?

jamesjjk:
@Pimm Correct using Mollie Connect.
As we written our own library to patch some of the problems, I currently do not have the exact error, but in principal it gives an error that the Order exists but the mode is wrong. I believe the testmode parameter is not being sent as a x-form header and encoded parameter but instead as a query parameter.

Pimm:
Could you try this on the new beta version?
`npm install @mollie/api-client@beta`

Regarding the library that you are using in-house: I would very much like to improve this library to the point where there is no more need for a third-party alternative. Would you be open to getting in touch so we can discuss this further?

Pimm:
This should be fixed in the current version of the library. Please create a new issue if you have reason to believe otherwise.

PR #148, opened 2020-05-20
Provide data in DELETE requests

Pimm:
This seems to work, and seems to be [in line with the PHP version](https://github.com/mollie/mollie-api-php/blob/f3ba2f3e776eb6b9ff0973e817282dd759c67d82/src/Endpoints/EndpointAbstract.php#L121).

Possibly related to #147.

### DIFF
--- a/src/resources/Resource.ts
+++ b/src/resources/Resource.ts
@@ -26,7 +26,7 @@ export default class Resource<R, T extends R> {
     get: (url: string, query?: Record<string, any>) => Promise<T>;
     list: (url: string, resourceName: string, query: Record<string, any>) => Promise<Omit<List<T>, 'nextPage' | 'previousPage'>>;
     patch: (url: string, data: any) => Promise<T>;
-    delete: <S extends T | true>(url: string) => Promise<S>;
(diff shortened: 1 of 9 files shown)


src/resources/Resource.ts
```diff
@@ -87,9 +87,9 @@ export default class Resource<R, T extends R> {
         }
         return this.injectPrototypes(response.data);
       },
-      delete: async <S extends T | true>(url: string): Promise<S> => {
+      delete: async <S extends T | true>(url: string, context: any): Promise<S> => {
```
vernondegoede, line 90: Does this need a default value?

Pimm: We could do `undefined` by default. I'm OK with that.

vernondegoede: It's also fine with leaving it like this.

Pimm: I thought about it a little more, and now I like the idea a little more. I just changed it.

Pimm:
@vernondegoede Could you retrigger Travis? I think it failed to notify GitHub of the results.

PR #158, opened 2020-07-07
Upgrade `kind-of` dependency to fix security vulnerability

vernondegoede:
This fixes https://github.com/advisories/GHSA-6c8f-qphg-qjgp.

Unfortunately, not all dependencies have released new versions that depend on the latest version of the `kind-of` dependency. Therefore, I suggest fixing this using selective dependency resolutions.

We should plan a moment to revisit all dependencies and upgrade the ones that released a patch. This will most likely allow us to clean up the other selective dependency resolutions as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
     "@babel/core": "^7.3.3",
     "babel-core": "7.0.0-bridge.0",
     "minimist": "^1.2.2",
-    "acorn": "^7.1.1"
+    "acorn": "^7.1.1",
+    "kind-of": "^6.0.3"
   }
 }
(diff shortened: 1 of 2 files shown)


CosmaTrix:
Instead of installing `kind-of` manually to bump its version, I'd use `Selective dependency resolutions`. That is a clean way to upgrade a peer-dependency without waiting for all the third-parties to upgrade.

https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

vernondegoede:
That's what I did @CosmaTrix

CosmaTrix:
Oh, right! I didn't expand the package.json and I thought those were `devDependencies`. 🤦

Issue #105, opened 2019-03-24
Client fails to import pem file on AWS lambda


danlutz:
Because of a known problem with the ```__dirname``` variable in serverless lambda functions, it is not possible to use the mollie-api-node module in a serverless environment.

See: https://github.com/serverless/serverless/issues/4778

A possible fix would be to inline the pem file as part of the build process of the module, or somehow work around using __dirname in /dist/mollie.js on line 35:
```js
options.httpsAgent = new https.Agent({
    cert: fs.readFileSync(path.resolve(__dirname, cert))
});
```

If you have any other solutions for this issue, I would be glad to hear about it

smonist:
i ran into the same problem yesterday - i fixed it by using [raw-loader](https://www.npmjs.com/package/raw-loader) and replacing the fs.readFileSync(path.resolve(__dirname, cert)) call with a require at the top of the file. its a quick workaround, if there are better solutions please let me know!

danlutz:
Nice solution! Did you fork the module for this?
Our solution was to just manually implement the apis we needed, but it would be much nicer to use the official client 👍

smonist:
nope, not yet. but i will propably create a fork tomorrow :)

TheDancingCode:
Having the same issue.

smonist:
i changed a few small things in their config files and finally got their module bundler to inline the .pem file
enjoy: https://github.com/mollie/mollie-api-node

kokeksibir:
If you use webpack (probably because you are developing typescript)  then here is my **workaround**:

1. Mollie calculates the path of certificate using `__dirname`. I used `string-replace-loader` to replace `__dirname` in mollie code with `process.env.LAMBDA_TASK_ROOT` 
```module: {
    rules: [
      ...
      {
        test: /mollie(.cjs|.esm)?\.js$/,
        loader: "string-replace-loader",
        options: {
          search: "__dirname",
          replace: "process.env.LAMBDA_TASK_ROOT",
          flags: "i"
        }
      }
    ]
  }```

2. Copy the certificate to root using `copy-webpack-plugin` plugin

```
  plugins: [
    new CopyPlugin([
      {
        from: "node_modules/@mollie/api-client/dist/cacert.pem",
        to: "./",
        flatten: true
      }
    ])
  ]
```

ndotmdot:
> i changed a few small things in their config files and finally got their module bundler to inline the .pem file
[...]

PR #162, opened 2020-08-17
Inlined certificates

Pimm:
1. The certificates are passed as `ca` instead of `cert`<sup>[1]</sup>.
2. The certificates are now bundled into the JavaScript files by rollup.js. This makes the library more portable, fixing #105.

What it looks like to me ‒ and I am not a security expert ‒ is that we are not using the certificates at this time. Instead, we [fall back](https://github.com/nodejs/node/blob/f5c0e282ccf98e17f295c11850649ad19a6fff51/lib/_tls_common.js#L105) on [the certificates included in Node.js](https://github.com/nodejs/node/blob/f5c0e282ccf98e17f295c11850649ad19a6fff51/src/node_root_certs.h).

With this PR, the certificates included in Node.js should be ignored in favour of the ones we provide.

Additionally, we are no longer loading the certificates file from the file system at runtime. This means we are no longer doing a `require('fs')`.

## To make our users not shoot themselves in the foot

There is a side-effect. In the current version, (mistakenly) integrating the library in a React app using create-react-app would cause this error:
```
TypeError: […].readFileSync is not a function
```
It breaks, because `require('fs')` doesn't provide the required API. However, this is the very line this PR removes…

Without any extra checks, with this PR merged a developer can call `createMollieClient` in a React app with no warnings/errors. (This is due to webpack providing a polyfill for [`https`](https://github.com/jhiesey/stream-http) and [`url`](https://github.com/defunctzombie/node-url). It seems webpack 5 will cease this behaviour, but the current version of webpack still has it.)

This PR includes an extra check, which detects polyfills for `https` ‒ and hopefully detects integrations in websites or apps that way.

## Thoughts

We could choose to rely on certificates included in Node.js (which we are doing now by accident) instead of providing our own list. This means giving up control and trusting Node.js to keep our users safe.

Alternatively, we could choose to only provide the certificate we need (i.e. USERTrust RSA Certification Authority).

[1]: https://node.readthedocs.io/en/latest/api/tls/#tlscreatesecurecontextdetails


Pimm:
Screenshot of error when integrating in React app

src/createMollieClient.ts
```diff
@@ -99,9 +99,18 @@ function createHttpClient({ apiKey, accessToken, versionStrings, ...axiosOptions
...
 
   axiosOptions.httpsAgent = new https.Agent({
-    cert: fs.readFileSync(path.resolve(__dirname, './cacert.pem'), 'utf8'),
+    ca: caCertificates,
   });
 
+  // Detect stub polyfills of https.Agent. Using a polyfill (such as https://github.com/jhiesey/stream-http) in itself
+  // is not an issue, although it may not respect the certificates and thereby reduce security somewhat. The purpose of
+  // this check is to catch cases where this library is integrated into a frontend app.
+  if (axiosOptions.httpsAgent?.requests == undefined) {
```
vernondegoede, line 108: Why not check for the presence of the `window` object?

vernondegoede:
I left 1 minor (nitpicky) comment. Thanks for thinking of that case.

Tested this locally, and it works great! Thanks for this nice PR 👏

src/createMollieClient.ts
Pimm: Another alternative would be checking `require('process').release.name`<sup>[1](https://nodejs.org/api/process.html#process_process_release)</sup>. I would prefer doing that over checking `window`. Polyfilling `window` is common (React Native does it, to name one).

I don't think there's a perfect solution. Ideally, we would check whether the library is running on a user's machine. But even Node.js could be running on a user's machine (think Electron).

Should we check `release.name` instead of `httpsAgent.requests`?

vernondegoede: Valid point on the polyfills – didn't think of that.

Let's go for `release.name` 👍

PR #166, opened 2020-09-03
Add subscriptions_payments api

andonimihai:
I added functionalities for this endpoint in NodeJS SDK: https://docs.mollie.com/reference/v2/subscriptions-api/list-subscriptions-payments . We need to use it for one of our clients and I really appreciate if you can review and release a new version ASAP.


Pimm (approved):
This is outstanding!

PR #167, opened 2020-09-06
apiEndpoint option

### LINKED ISSUE #164: Allow changing API endpoint
We would like to use the Node API client to run tests in our CI. This would require us to change the API endpoint.

Would it be possible to expose a method similar to the PHP API client's `setApiEndpoint()`? The PHP source can be found here: https://github.com/mollie/mollie-api-php/blob/master/src/MollieApiClient.php#L328

Pimm:
Added `apiEndpoint` property to the options passed to `createMollieClient`.

I did not add any documentation, as I feel this feature doesn't require any extra discoverability. This is a feature which will not be of interest to the average user.

Fixes #164.


dsempel:
Looks good to me!

edorivai (approved):
LGTM

PR #170, opened 2020-10-14
stringifyQuery fix

### LINKED ISSUE #165: [BUG][methods]: breaking change
I just updated with the last version, I noticed a bug (it's not in the changelog) on the `methods` api.

I used to call the method API like this :

`mollieClient.methods.all({ include: 'issuers', amount: { value: '100.00', currency: 'EUR' })`

I can't use the `amount` field in the query anymore it throws `The value is missing from the amount` even if the value is correct.

Pimm:
In commit 886de38650d120e0be2a85e97e676e5cf9d87597, we switched from [an external alternative](https://www.npmjs.com/package/qs) to [Node.js' querystring module](https://nodejs.org/api/querystring.html).

Unbeknownst to me, said external alternative has special logic for nested objects on which we rely [in a specific endpoint](https://docs.mollie.com/reference/v2/methods-api/list-all-methods). This commit adds that special logic to our codebase.

Additionally, I added a test to verify the behaviour of `stringifyQuery` with objects.

Fixes #165.

### DIFF
--- a/src/resources/Resource.ts
+++ b/src/resources/Resource.ts
@@ -15,7 +15,9 @@ function stringifyQuery(input: Record<string, any>): string {
     entries.reduce<Record<string, any>>((result, [key, value]) => {
       if (Array.isArray(value)) {
         result[key] = value.join();
-      } /* if (Array.isArray(value) == false) */ else {
+      } else if (/* Array.isArray(value) == false && */ typeof value == 'object') {
+        getEntries(value).forEach(([innerKey, innerValue]) => (result[`${key}[${innerKey}]`] = innerValue));
+      } /* if (typeof value != 'object') */ else {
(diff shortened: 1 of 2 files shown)


dsempel (approved):
Bug fix ✅ 
Unit test ✅ 
Nothing to add, except that I'm wondering if there might be more endpoints with nested query params. Did you check? It's theoretically possible to have even more deeply nested objects.

Pimm:
Correct. This change only fixes this specific scenario: a first level nested object with primitive values (since it doesn't even handle arrays within such an object correctly).

From what I can tell ‒ at least by looking at our type definitions ‒ this is the only endpoint where we have something like this. So we should be OK. If it turns out we need more advanced query strings, I'll introduce smarter code.

PR #172, opened 2020-10-14
Pass include (payments.create) in the query string instead of the body

### LINKED ISSUE #168: `include` should be query string parameter instead of body, create payment API
the field `include` listed in body parameters, but API expects that to be part of the query string.

https://github.com/mollie/mollie-api-node/blob/d05abf261ac4e68f9e7bd7325d8c8155a3a928ea/src/resources/payments/parameters.ts#L44

sending `include` via body causes 

```
ApiError: Non-existent body parameter "include" for this API call. Did you mean: "locale"?
```

ref: https://docs.mollie.com/reference/v2/payments-api/create-payment#qr-codes

Pimm:
[According to the documentation](https://docs.mollie.com/reference/v2/payments-api/create-payment#qr-codes), the `include` property should be included in the query string instead of the body.

Fixes #168.

### DIFF
--- a/src/resources/payments/PaymentsResource.ts
+++ b/src/resources/payments/PaymentsResource.ts
@@ -1,5 +1,5 @@
 import { CancelParameters, CreateParameters, GetParameters, ListParameters, UpdateParameters } from './parameters';
-import { PaymentData } from '../../data/payments/data';
+import { PaymentData, PaymentInclude } from '../../data/payments/data';
 import ApiError from '../../errors/ApiError';
 import Callback from '../../types/Callback';
 import List from '../../data/list/List';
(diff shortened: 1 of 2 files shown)


src/resources/payments/PaymentsResource.ts
```diff
@@ -55,7 +55,12 @@ export default class PaymentsResource extends Resource<PaymentData, Payment> {
   public create(parameters: CreateParameters, callback: Callback<Payment>): void;
   public create(parameters: CreateParameters) {
     if (renege(this, this.create, ...arguments)) return;
-    return this.network.post(this.getResourceUrl(), parameters);
+    const { include, ...data } = parameters;
+    let query: { include: PaymentInclude[] | PaymentInclude } | undefined = undefined;
+    if (include != undefined) {
+      query = { include };
+    }
```
dsempel, line 62: I would write this a little bit different, because I feel that's more readable and makes nice use of type inference. Although I'll leave it up to you.
`const query = include ? { include } : undefined;`

dsempel (approved):

src/resources/payments/PaymentsResource.ts
Pimm: The reason I didn't, is because I don't want to silently drop the parameter if it is (accidentally) falsy.

I'll change it to this:
```javascript
const query = include != undefined ? { include } : undefined;
```
That's probably the cleanest solution.

dsempel: You need to have trust in the types 😄

PR #173, opened 2020-10-14
Add parameters to orders.update

### LINKED ISSUE #169: Redirect url including order id
I am using mollie in a serverless environment and I am running into a problem:

The flow I am aiming at:

1. Customer creates order and gets the checkout url from my endpoint
2. Customer completes or aborts payment
3. Customer gets redirected to the redirectUrl
4. The status page computes the order status by fetching the order from mollie (so no Database linking orderNumber to order id on my side)
5. Serverless webhooks manage the emails, the problem does not exist here because mollie provides the full order object

There is no way to make the `redirectUrl` include the order id. This would not be a problem if I keep track of the order in a database, but in my usecase the communication is stateless and order confirmation primarily uses emails.

Workarounds I have tried so far with no success are:

- Using my generated `orderNumber`, problem here is that I can't query a order by that id later.
- Creating the order with a generic redirect url and updating the order with a new `redirectUrl` before redirecting the customer to checkout. The problem here is that the node client does not allow the `redirectUrl` in the update object (the api seems to support it)

[The docs](https://docs.mollie.com/reference/v2/orders-api/update-order) suggest that updating the `redirectUrl` is allowed on uncompleted orders. Should this be reflected in the node-client?

I have just switched back to this client from my own implementation because of the serverless certificate fix and I want to avoid moving away from it again.

Pimm:
[According to the documentation](https://docs.mollie.com/reference/v2/orders-api/update-order), it is possible to update the redirect and webhook URLs for orders. Therefore, the associated properties are now included in the parameters type.

Fixes #169.

### DIFF
--- a/src/resources/orders/parameters.ts
+++ b/src/resources/orders/parameters.ts
@@ -47,7 +47,7 @@ export interface GetParameters {
   embed?: OrderEmbed[];
 }
 
-export type UpdateParameters = PickOptional<OrderData, 'billingAddress' | 'shippingAddress'> & {
+export type UpdateParameters = PickOptional<OrderData, 'billingAddress' | 'shippingAddress' | 'redirectUrl' | 'webhookUrl'> & {
   orderNumber?: string;
   testmode?: boolean;
 };


dsempel (approved):
So I guess this was already possible to do, but the types were just not up to date?

Pimm:
Exactly. It doesn't seem like this was added recently, either. I think this has simply been missing from the types.

PR #176, opened 2020-10-26
Add profileId and testmode to order creation parameters

DeMoorJasper:
This PR adds testmode and profileId to CreateParameters for Order.

To ensure this can be used when using Mollie Connect if you're relying on TypeScript to ensure type safety.


Pimm (approved):
Thank you, Jasper!

PR #180, opened 2020-10-29
Lines is optional in the shipments api

DeMoorJasper:
Lines should be optional in the shipments api (according to the API docs)


Pimm (approved):
Thank you, Jasper!

PR #181, opened 2020-11-19
Explain how to get refunds in getting payment

PanMan:
this was documented in the API doc but not how to use this in the node lib.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,19 @@ mollieClient.payments.get(payment.id)
   });
 ```
 
+
+### Retrieve a payment with refunds:
+
+```javascript
+mollieClient.payments.get(payment.id, { embed: "refunds" })
+  .then(payment => {
+    // E.g. check if the payment.isPaid()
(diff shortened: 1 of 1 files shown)


Pimm:
Thanks, PanMan!

The read-me is not meant to contain an exhaustive list of examples of everything one can do with the library. I do, however, agree that we could have better documentation on how to get embedded resources. You've probably opened this PR because figuring this out by yourself was unpleasant. That proves that there is room for improvement. The [get payments documentation](https://docs.mollie.com/reference/v2/payments-api/get-payment) currently reads:

> This endpoint also allows for embedding additional information by appending the following values via the `embed` query string parameter.

That is kind of odd when using any of the libraries (the Node.js one or otherwise).

So I agree that we can communicate this better, but I think the documentation ([docs.mollie.com](https://docs.mollie.com/reference/v2/payments-api/create-payment)) is the place for it; not the read-me.

PR #188, opened 2021-01-11
Refactor networking

Pimm:
Before this refactor, the `Resource` class was responsible for some of the networking logic. After it, this responsibility is split into a `NetworkClient` class (shared by all resources) and a `TransformingNetworkClient` (one for each resource).

This simplifies both the `Resource` class and the `createMollieClient.ts` module.

The `Resource` (base)class is now a candidate for removal. It only contains the `injectPaginationHelpers` helper function. The `ParentedResource` (base)class is candidate for removal as well, as it only exists to facilitate `withParent` (which has been deprecated since 2.0.0).

### DIFF
--- a/src/Options.ts
+++ b/src/Options.ts
@@ -1,0 +1,33 @@
+import { AxiosRequestConfig } from 'axios';
+import Xor from './types/Xor';
+
+type Options = Xor<
+  {
+    /**
+     * The Mollie API key, starting with `'test_'` or `'live_'`.
+     */
+    apiKey: string;
+  },
+  {
+    /**
+     * OAuth access token, starting with `'access_''.
+     */
+    accessToken: string;
+  }
+> & {
+  /**
+   * One or an array of version strings of the software you are using, such as `'RockenbergCommerce/3.1.12'`.
+   */
+  versionStrings?: string | string[];
+  /**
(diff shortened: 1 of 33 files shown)


src/Options.ts
```diff
@@ -0,0 +1,33 @@
...
+> & {
+  /**
+   * One or an array of version strings of the software you are using, such as `'RockenbergCommerce/3.1.12'`.
+   */
+  versionStrings?: string | string[];
+  /**
+   * The headers set in the requests sent to the Mollie API. `Authorization`, `User-Agent`, `Accept-Encoding`, and
+   * `Content-Type` are set by this library directly, setting them here has no effect.
+   */
+  headers?: Record<string, string>;
```
edorivai, line 26: Why did you manually add the `headers` field here, instead of `Pick<>`'ing from `AxiosRequestConfig`?

Pimm: `AxiosRequestConfig` defines `headers` as `any`, for some reason. `Record` is more specific, and it means the cloning is more predictable (see the constructor of `NetworkClient`).

edorivai: Ok, makes sense. LGTM

edorivai (approved):

PR #189, opened 2021-01-12
Update dependencies

Pimm:
This should make #187 and #184 obsolete.

We are currently stuck at jest 24.9.0, because 25.1.0 drops Node.js 6 support.


edorivai (approved):
LGTM

PR #193, opened 2021-01-22
Support applepay validation

hobadams:
I've tried to add support for the ApplePay session method - https://docs.mollie.com/reference/v2/wallets-api/request-apple-pay-payment-session

I've not been able to test because this method is not available in test mode. 

I'd love some feedback if you're able to provide it

### DIFF
--- a/src/createMollieClient.ts
+++ b/src/createMollieClient.ts
@@ -99,7 +99,7 @@ export default function createMollieClient(options: Options) {
     // Onboarding API
     onboarding: new OnboardingResource(networkClient),
 
-    // ApplePay API
+    // Apple Pay API
     applePay: new ApplePayResource(networkClient),
   };
   /* eslint-enable @typescript-eslint/camelcase */
--- a/src/createMollieClient.ts
+++ b/src/createMollieClient.ts
@@ -5,6 +5,7 @@ import NetworkClient from './NetworkClient';
 import Options from './Options';
 
 // Resources
+import ApplePaySessionResource from './resources/applePaySession/ApplePaySessionResource';
 import ChargebacksResource from './resources/chargebacks/ChargebacksResource';
 import CustomersMandatesResource from './resources/customers/mandates/CustomersMandatesResource';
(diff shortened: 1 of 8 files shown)


hobadams:
@Pimm - Are you the best person to request a review from?

hobadams:
@Pimm - Thanks so much for looking at this so quickly.

Pimm:
Thank you for the contribution!

I changed your PR in two ways:
* I renamed the method to `requestPaymentSession` to match the documentation, and
* I made `ApplePaySessionData` not inherit from `Model`, as the object does not actually have an `id` and `resource` property like most objects returned by the API do.

It is unfortunate that we can't test this in the way we test the other endpoints.

hobadams:
Awesome, thanks so much @Pimm. Do you think this will make it's way into your next release?

Pimm:
Yes!

Pimm (approved):

edorivai (approved):
Changes generally look good to me. I would advise to issue a prerelease that @hobadams can start using in their app, given that we don't have the ability to test applepay in a test environment.

@hobadams would you be willing to test this out and report back with results after a while?

hobadams:
@edorivai  - good idea. I'll be using it with our Mollie / Deity (https://deity.io/) integration so it should get heavily tested.

hobadams:
@Pimm @edorivai - I've just tested the merged code using production Mollie credentials and it's returning the data I expected. I've still got a little way before I have a full integration working in my app to show you but this works well.

Thanks!

Pimm:
To hopefully make your job slightly easier, you can now use the version with the `beta` dist tag (using `npm install @mollie/api-client@beta`). It reflects master at this point in time, and includes your contribution.

Should you run into any issues while integrating Mollie ‒ including anything not Apple Pay related ‒ please let me know.

hobadams:
@Pimm I've put live ApplePay transactions through Mollie using the beta branch so I can confirm it works :-)

PR #195, opened 2021-01-25
Add documentation from docs.mollie.com

Pimm:
I copied documentation from [docs.mollie.com](https://docs.mollie.com) into the source code using my new experimental tool [atsee](https://github.com/Pimm/atsee).

Before this pull request, this library had some of the documentation on docs.mollie.com copied into the source code by hand, while some methods/properties were "naked": lacking any documentation.

Naked methods/properties are unfortunate, as developers might have to take a trip out of their IDE to look something up. However, copying over documentation by hand has the risk of the copy in the source code becoming outdated. (Copying over things by hand is not being a particularly fun chore, either).

This might be a healthy compromise. The copied-over documentation should improve the user experience, while the automation of the copying process makes it easy to keep the source code in sync with docs.mollie.com.

### DIFF
--- a/src/data/chargebacks/Chargeback.ts
+++ b/src/data/chargebacks/Chargeback.ts
@@ -63,7 +63,17 @@ type Chargeback = Seal<
 export default Chargeback;
 
 export interface ChargebackLinks extends Links {
+  /**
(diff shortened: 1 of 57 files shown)


edorivai (approved):
This is really good stuff!

PR #196, opened 2021-01-28
Replace Node.js URL API for WHATWG URL API.

Pimm:
The [Node.js URL API](https://nodejs.org/api/url.html#url_legacy_urlobject) was deprecated in Node.js 11, while the [WHATWG counterpart](https://url.spec.whatwg.org/) is supported since Node.js 7 and 6.13.

Our lowest supported Node.js version is currently 6.14.

### DIFF
--- a/src/resources/Resource.ts
+++ b/src/resources/Resource.ts
@@ -11,7 +11,8 @@ import TransformingNetworkClient from '../TransformingNetworkClient';
  */
 function parseQueryInUrl(url: string) {
   const result: Record<string, string> = {};
-  return new URL(url).searchParams.forEach((value, key) => result[key] = value), result;
+  new URL(url).searchParams.forEach((value, key) => (result[key] = value));
+  return result;
 }
 export default class Resource<R, T extends R> {
   constructor(protected readonly networkClient: TransformingNetworkClient<R, T>) {}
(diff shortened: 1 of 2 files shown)


src/resources/Resource.ts
```diff
@@ -1,8 +1,18 @@
...
+/**
+ * Returns the parsed search parameters from the passed URL. For example: `'https://example.com?id=5'` is converted to
+ * `{ id: 5 }` (and `'https://example.com'` is converted to `{}`).
+ *
+ * If multiple parameters have the same key (`'https://example.com?id=5&id=6'`), exactly one of them will be
+ * represented in the returned object.
+ */
+function parseQueryInUrl(url: string) {
+  const result: Record<string, string> = {};
+  return new URL(url).searchParams.forEach((value, key) => result[key] = value), result;
```
edorivai, line 14: Code style nit. I think splitting this expression over 2 lines will be more commonly understood.

```suggestion
  new URL(url).searchParams.forEach((value, key) => result[key] = value);
  return result;
```

edorivai (changes requested):

src/resources/Resource.ts
Pimm: Sure!

edorivai (approved):
LGTM!

PR #199, opened 2021-03-12
Add applePayPaymentToken to payment parameters

Timollie:
Make it possible to create an apple pay payment when using typescript.
This will otherwise throw a "not assignable" error.


Pimm (approved):
Perfect. Thanks!

PR #200, opened 2021-03-15
Update documentation from docs.mollie.com.

Pimm:
This reflects [the discontinuation of ING Home'Pay](https://docs.mollie.com/changelog/v2/changelog#february-2021) in the library.

### DIFF
--- a/src/data/payments/data.ts
+++ b/src/data/payments/data.ts
@@ -115,8 +115,8 @@ export interface PaymentData extends Model<'payment'> {
    *
    * If the payment is only partially paid with a gift card, the method remains `giftcard`.
    *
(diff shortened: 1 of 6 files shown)


src/data/payments/data.ts
```diff
@@ -558,23 +558,8 @@ export interface IdealDetails {
 }
 
 export interface IngHomePayDetails {
```
edorivai, line 560: We debated changing any interfaces for deprecated payment methods to `any` - this would reflect Mollie's policy of completely removing the documentation pages for these methods.

edorivai: I'll leave the final call to you on whether you want to do this.

edorivai (approved):

src/data/payments/data.ts
Pimm: I'm putting that on the todo list. I would like to experiment with the type of the `details` property being dependent on the value of `method`. Those two are related and should maybe be done in one go.

PR #211, opened 2021-06-27
Add Paypal Support

lindit:
Updated the CreateParameters for mandates to support PayPal option.
Like here: 
https://docs.mollie.com/reference/v2/mandates-api/create-mandate

### DIFF
--- a/src/resources/customers/mandates/parameters.ts
+++ b/src/resources/customers/mandates/parameters.ts
@@ -31,6 +31,14 @@ export type CreateParameters = ContextParameters &
      *
      * @see https://docs.mollie.com/reference/v2/mandates-api/create-mandate?path=signatureDate#parameters
      */
+
+    /**
+     * The consumer’s email address. Required for paypal mandates
+     *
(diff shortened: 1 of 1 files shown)


lindit:
Could you tell me after the merge when this will get released and what version that'll be?

Pimm:
Thank you, Lindit. I will look into this, and keep you up-to-date.

Please ‒ however ‒ do not wait for a new version to be published before you continue development. You can already send a `consumerEmail` through the API today. If you're using plain JavaScript, you should have no issues. If you are using TypeScript, you might need a `@ts-ignore` or an `as any` somewhere (until the new version rolls out). There is never a need to be blocked by outdated type definitions. If you need assistance, please let me know.

Pimm (approved):

Pimm:
Hi @lindit,

It took way longer than expected, but your PR has landed. You can use it today by installing the _beta_ tag: `npm install @mollie/api-client@beta`. Should you have any issue with the beta release, please let us know.

PR #214, opened 2021-08-04
Updated Travis CI configuration and dependencies

Pimm:
* Updated Travis CI configuration, including [the removal of `sudo: false`](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration#timeline---its-happening-fast).
* Updated dependencies (only dev dependencies were outdated).
* Fixed [test](//github.com/Pimm/mollie-api-node/commit/77068560bfb63b407b93687f1fd0a08372fa39bd).
* Removed stray publish logic from Travis CI configuration, as we aren't using that.
* Removed stray Husky dependency.


dsempel (approved):
Looks good to me, just some infra changes. Although I'm wondering why that test didn't fail before..

Pimm:
> Although I'm wondering why that test didn't fail before..

The `expect` is inside `if (!orders.length)`. This test creates an order only if there are none (and only checks the ID if the order was created).

Tests previously ran with a different API key. An API key for an account which already had at least one order.

I do plan on reviewing all of these tests in the near future. They should behave consistent ‒ regardless of the payments/orders/customers which exist or do not exist on the test account.

PR #233, opened 2021-11-05
Deprecate predictable helper methods

Pimm:
Deprecated methods/properties such as:
* `payment.isOpen()` (synonymous with `payment.status == PaymentStatus.open`)
* `list.count` (synonymous with `list.length`)
* `error.getMessage()` (synonymous with `error.message`)

### Cons

1. Removing these methods/properties will be a breaking change.
2. Some of these methods/properties also exist in the PHP client, therefore removing them will also widen the gap between the two clients.

### Pros
1. Consumers of an API expect every part of it to be significant. When a consumer sees that a list has both a `length` and a `count`, they might wonder "What is the difference between these two? Perhaps `length` is the number of elements in this array, where `count` is the total number of elements available?" Methods/properties with such simple synonyms can cause confusion.
2. Having a narrower API enhances discoverability. Without `isOpen`, `isAuthorized`, `isExpired`, `isFailed`, `isPending`, `hasSequenceTypeFirst`, and `hasSequenceTypeRecurring`, it's easier to find `getCheckoutUrl` and `getRefunds`. It's easier to find the needle if the haystack is smaller.
3. Having fewer roads leading to Rome promotes consistency in the code which uses this library.

### DIFF
--- a/src/createMollieClient.ts
+++ b/src/createMollieClient.ts
@@ -158,3 +158,5 @@ export { OrderLineType } from './data/orders/orderlines/OrderLine';
 export { PaymentEmbed, PaymentStatus } from './data/payments/data';
 export { RefundEmbed, RefundStatus } from './data/refunds/data';
 export { SubscriptionStatus } from './data/subscription/data';
+export { ProfileStatus } from './data/profiles/data';
+export { OnboardingStatus } from './data/onboarding/data';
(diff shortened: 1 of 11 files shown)


edorivai (approved):
LGTM

dsempel:
I agree this is better.

PR #235, opened 2021-11-12
Implement iteration API, in addition to pagination API

Pimm:
### Rationale
The Mollie API provides paginated endpoints, which are great if a developer using this library is implementing a paginated dashboard.

Some developers ‒ myself included ‒ (ab)use these paginated endpoints and the `nextPage` method to apply logic to _all_ values in a set.
```javascript
mollieClient.payments.page()
.then(payments => {
	// Apply logic.
	return payments.nextPage();
})
.then(payments => {
	// Apply logic.
	return payments.nextPage();
});
```
This is inconvenient to the developer, because the `nextPage` pattern above requires a lot of boilerplate code; and depending on the exact code might be inefficient in terms of memory usage as well as load on the Mollie API.

### Basic
This pull requests implements an API which lives alongside the paginated API, and is designed for this case:
```javascript
for await (let payment of mollieClient.payments.iterate()) {
	// Apply logic.
}
```

### Advanced
There is a (currently stage 2) [proposal to add helpers to iterators](https://github.com/tc39/proposal-iterator-helpers). In anticipation of the inclusion of these helpers, some of them are polyfilled in our library.

This powers the following snippets:
```javascript
// Find the 10 latest euro payments over €100.00.
const iterator = mollieClient.payments.iterate()
	.filter(({ amount }) => amount.currency == 'EUR' && parseFloat(amount.value) > 100)
	.take(10);
for await (let payment of iterator) {
	// Apply logic.
}
```
```javascript
// Find a payment by description.
const payment = await mollieClient.payments.iterate()
	.find(({ description }) => description == 'Transaction #b0e547a7');
```

With a bit of luck, we can discard our polyfill if and when the proposal is accepted without any breaking changes.


edorivai (commented):
Only comment I have is that it might be a good idea to write some unit tests for the `HelpfulIterator`, what do you think?

Pimm:
Thanks, @edorivai. `iteration/iteration.test.ts` is essentially a test for `HelpfulIterator`. It doesn't test it directly like a unit test would, but it does cover all of the methods (`forEach`, `filter`, et cetera) except for `map` ‒ which `TransformingNetworkClient` uses and is therefore automatically covered.

What could be a good idea, though, is to add a test which iterates over several pages. `iteration/iteration.test.ts` only covers a handful of payments, so they're all on the same page.

tests/iteration/iteration.test.ts
```diff
@@ -0,0 +1,133 @@
+import createMollieClient, { ApiMode, MollieClient, Payment, PaymentMethod } from '../..';
+import axios from 'axios';
+import nock from 'nock';
+import { setupRecorder } from 'nock-record';
+import { run } from 'ruply';
+import getAccessToken from '../getAccessToken';
+
+const mockNetwork = true;
```
edorivai, line 8: Let's add a comment as to what this does

Pimm: Can do!

edorivai:
Oh yeah, that makes sense. These tests should cover it then

edorivai (approved):

PR #237, opened 2021-11-15
Lodash removal

Pimm:
Removed Lodash as a dependency. While depending on Lodash is not at all problematic, our usage did not warrant the extra dependency.

### DIFF
--- a/src/errors/ApiError.ts
+++ b/src/errors/ApiError.ts
@@ -1,5 +1,4 @@
 import { AxiosResponse } from 'axios';
-import { cloneDeep, get, has } from 'lodash';
 
 import { MollieApiErrorLinks, Url } from '../data/global';
 import Maybe from '../types/Maybe';
@@ -67,7 +66,7 @@ export default class ApiError extends Error {
    * @since 3.0.0
    */
   public hasLink(key: string): boolean {
-    return has(this.links, key);
+    return this.links !== undefined && key in this.links;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/errors/ApiError.ts
```diff
@@ -97,11 +96,11 @@ export default class ApiError extends Error {
...
-      get(response, 'data.detail', 'Received an error without a message'),
-      get(response, 'data.title'),
-      get(response, 'data.status'),
-      get(response, 'data.field'),
-      cloneDeep(get(response, 'data._links')),
+      response.data.detail ?? 'Received an error without a message',
+      response.data.title,
+      response.data.status,
+      response.data.field,
+      response.data._links,
```
dsempel, line 103: `data._links` is no longer cloned, but I guess that's ok? Theoretically someone could mutate a `Url` obtained through `getLink`.

dsempel (approved):
Good that you were able to get rid of Lodash!

src/errors/ApiError.ts
Pimm: The cloning behaviour was introduced in e24d9e531983596ebfdc12a035164df0efe5d9e7 (February '19), I'm not entirely sure for what reason.

`response.data` is a fresh object coming from axios, and it is only consumed here. So whether the original is used or it is cloned first doesn't make any difference I can imagine.

Perhaps the idea was indeed to prevent someone from doing `getLink` and then altering the returned object. However, to prevent that, the cloning would have to happen inside `getLink``.

For as far as I can tell, not cloning does not result in a noticeable or measurable change.

PR #240, opened 2021-11-17
Deprecate payment.getPaymentUrl

Pimm:
Deprecated `payment.getPaymentUrl()`.

A breaking change, but:
 * As this is synonymous with `payment.getCheckoutUrl()`, it is a potential cause for confusion.
 * Since the  PHP client only has `getCheckoutUrl`, this removal make this library more consistent with the PHP client.

### DIFF
--- a/src/data/payments/PaymentHelper.ts
+++ b/src/data/payments/PaymentHelper.ts
@@ -75,6 +75,8 @@ export default class PaymentHelper extends Helper<PaymentData, Payment> {
 
   /**
    * Returns the URL the customer should visit to make the payment. This is to where you should redirect the consumer.
+   *
+   * @deprecated Use `payment.getCheckoutUrl()` instead.
    */
   public getPaymentUrl(): Nullable<string> {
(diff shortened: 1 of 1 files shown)


src/data/payments/PaymentHelper.ts
```diff
@@ -75,6 +75,8 @@ export default class PaymentHelper extends Helper<PaymentData, Payment> {
 
   /**
    * Returns the URL the customer should visit to make the payment. This is to where you should redirect the consumer.
+   *
+   * @deprecated Use `payment.getCheckoutUrl()` instead.
    */
   public getPaymentUrl(): Nullable<string> {
     return this.links.checkout?.href ?? null;
```
dsempel, line 82: Since the method is so simple it wouldn't bring you much, but you could even do:
```suggestion
  public getPaymentUrl(): Nullable<string> {
    return this.getCheckoutUrl();
```

dsempel (approved):
Makes sense.

PR #247, opened 2021-11-25
Made list the preferred method for endpoints which don't paginate

Pimm:
`page` was selected as the preferred method name over `list` and `all` in #234. This PR re-introduces `list` as the preferred method name _for endpoints which don't paginate_.

Those endpoints are:
* [List methods](https://docs.mollie.com/reference/v2/methods-api/list-methods)
* [List shipments](https://docs.mollie.com/reference/v2/shipments-api/list-shipments) (for which the documentation does not explicitly mention that it does not paginate, but according to my tests it doesn't)
* [List permissions](https://docs.mollie.com/reference/v2/permissions-api/list-permissions)

The `iterate` method does not exist for these endpoints, either.

### DIFF
--- a/src/binders/methods/MethodsBinder.ts
+++ b/src/binders/methods/MethodsBinder.ts
@@ -31,7 +31,7 @@ export default class MethodsBinder extends Binder<MethodData, Method> {
    * @deprecated Use `page` instead.
    * @see https://docs.mollie.com/reference/v2/methods-api/list-methods
    */
-  public all: MethodsBinder['page'] = this.page;
+  public all: MethodsBinder['list'] = this.list;
   /**
    * Retrieve all enabled payment methods. The results are not paginated.
    *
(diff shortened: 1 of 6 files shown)


src/binders/methods/MethodsBinder.ts
```diff
@@ -31,7 +31,7 @@ export default class MethodsBinder extends Binder<MethodData, Method> {
    * @deprecated Use `page` instead.
```
dsempel, line 31: You forgot this one

src/binders/orders/shipments/OrderShipmentsBinder.ts
```diff
@@ -24,15 +24,15 @@ export default class OrderShipmentsBinder extends InnerBinder<ShipmentData, Ship
    * @deprecated Use `page` instead.
```
dsempel, line 24: And this one

dsempel (approved):
Makes sense to me, page is weird if there's only one page.

PR #251, opened 2021-12-13
Update order and payment types

Pimm:
This PR includes a few updates to the types for orders and payments. Most notably, the `payment` field for the order creation parameters is more accurate[[doc]](https://docs.mollie.com/reference/v2/orders-api/create-order#payment-parameters).

Additionally, this PR bumps the version to 3.6.0-beta.2, as I intend to publish immediately (on the _beta_ tag).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mollie/api-client",
-  "version": "3.6.0-beta.1",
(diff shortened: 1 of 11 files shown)


dsempel (approved):
Some nice improvements!

PR #283, opened 2022-07-20
Throttle iteration API

### LINKED ISSUE #271: Discourage developers from requesting large sets of data
### Background
Since version 2.0.0 of this library, calling a paginated endpoint returns an array with added `nextPageCursor` and `previousPageCursor` properties. These properties are great for implementing a paginated view.

The array also has added `nextPage` and `previousPage` methods ([similar to `next` and `previous` in the PHP client](https://github.com/mollie/mollie-api-php/blob/647ae292ffeae60e29be14688c2259038a8cacbe/src/Resources/CursorCollection.php#L37)). The only purpose of these methods is to request large sets of data ‒ sets which span across multiple pages.

```javascript
const firstPage = await mollieClient.payments.page();
const secondPage = await firstPage.nextPage();
const thirdPage = await secondPage.nextPage();
```

Why would a developer request large sets of data? They might need filtered data (all Italian customers); or they might be looking for a specific value but don't know the ID (getting a payment by its foreign ID in the metadata).

Ideally, the Mollie API would cover these cases. The clients would hand off the filtering/searching work to the Mollie API, which would send back the result. However, the Mollie API currently doesn't do any filtering or searching (not even trivial examples), and besides that it might never cover all use cases.

In version 3.6.0 of this library, we [introduced the iteration API](https://github.com/mollie/mollie-api-node/pull/235) (`iterate` method), which is essentially the [JavaScript-native alternative](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/for-await...of) for `nextPage` (and `previousPage`).

Usage:
```javascript
// Find the 10 latest euro payments over €100.00.
const iterator = mollieClient.payments.iterate()
	.filter(({ amount }) => amount.currency == 'EUR' && parseFloat(amount.value) > 100)
	.take(10);
for await (let payment of iterator) {
	// Apply logic.
}
```
```javascript
// List all Italian customers.
const iterator = mollieClient.customers.iterate()
	.filter(({ locale }) => locale == 'it_IT');
for await (let customer of iterator) {
	// Apply logic.
}
```
```javascript
// Find out whether a certain customer has spent more than €1000 in a single payment.
const bigSpender = await mollieClient.customerPayments.iterate({ customerId: 'cst_test' })
	.some(({ amount }) => amount.currency == 'EUR' && parseFloat(amount.value) > 1000);
```

### The problem
Mollie has experienced incidents where developers were making so many requests in rapid succession that the API was having performance and reliability issues.

Developers can write this horrible code since 2018:
```javascript
// ❌ Bad! Never do this.
var payments = await mollieClient.payments.page();
do {
	console.log(payments);
	payments = await payments.nextPage?.();
} while (payments);
```

With the iteration API, they can write:
```javascript
// ❌ Bad! Never do this.
for await (let payment of mollieClient.payments.iterate()) {
	console.log(payment);
}
```
(Note that there is no `break` keyword in the for-loop.)

The version with the iteration API is two lines shorter, and is less explicit about how many requests it makes (there is no explicit `nextPage` call).

People within Mollie have expressed concern that the problem they're already having ‒ developers making many requests in rapid succession ‒ may worsen now that we have the iteration API.

### Caveats
Developers are using this client voluntarily. They can opt-out of using it at any time, and just use curl to communicate with the Mollie API instead. Nothing we can do in the client prevents a developer from performing a ~~D~~DoS attack on the Mollie API.

What we can do, is make developers aware of the stress they put on the Mollie API. And perhaps we can make it easier to get the job done within a reasonable number of requests.

Pimm:
As described in [this issue](//github.com/mollie/mollie-api-node/issues/271), developers should not request large sets of data from the Mollie API in most scenarios. This PR aims to discourage doing so, by introducing throttling.

If a developer (and thereby a merchant) requests a large set of data, the throttling will spread out the pressure on the Mollie API over a larger period of time. But additionally, it might send a signal to the developer, asking "hey, this thing you're doing, it's costly, is there no better way?"

(Closes #271)

### Interface
In this implementation, the developer is in control of the throttling:
```javascript
for await (let payment of mollieClient.payments.iterate({ valuesPerMinute: 5000 })) {
  […]
}
```
Developers are using this client voluntarily. They can opt-out of using it at any time, and just use curl to communicate with the Mollie API instead. This is why I feel "forcing" a certain amount of throttling is not productive. If a developer chooses to set the `valuesPerMinute` to `Number.POSITIVE_INFINITY`, that's a conscious decision on their part.

(`valuesPerMinute` is optional, making this change backward compatible.)

### Better
A safer form of throttling ‒ in the sense that it's more difficult for developers to circumvent ‒ would be one that is applied on the Mollie API itself rather than in the client(s). When such throttling is introduced, it can be removed from the client.

### Best
Neither spreading the requests over a longer period of time, nor asking the developer to find a way to make fewer request is an ideal solution. The real solution is making _more efficient_ requests, by supporting filtering and searching in the Mollie API itself.

Once the Mollie API supports filtering payments by currency, for instance, we can expand the iteration API:
```javascript
// Find the 10 latest euro payments.
const iterator = mollieClient.payments.iterate()
	.filter({ currency: 'EUR' })
	.take(10);
for await (let payment of iterator) {
	// Apply logic.
}
```
This would guarantee _one single_ call to the Mollie API (`v2/payments?currency=EUR&limit=10`).


src/communication/NetworkClient.ts
```diff
@@ -161,28 +145,76 @@ export default class NetworkClient {
...
+            const response = await axiosInstance.get(url).catch(throwApiError);
+            try {
+              /* eslint-disable-next-line no-var */
+              var { _embedded: embedded, _links: links } = response.data;
+            } catch (error) {
+              throw new ApiError('Received unexpected response from the server');
+            }
+            // Yield the values on the page.
+            const values = embedded[binderName] as R[];
+            yield* values;
```
janpaepke, line 193: This is more of a question for my understanding, because I probably have a thinking error somewhere:
Shouldn't `throttler.tally` be called before `yield`?

From what I understand the generator function will be halted at `yield` on every call of `.next()`.
So currently it would halt before tallying the values. It would only tally on the subsequent `next()`  call (after depleting values). However if enough time has passed between the previous call and this one already, the values wouldn't have to be throttled anymore, right? 🤔

Pimm: I'm not 100% sure whether I'm following your reasoning, but let's see.

`tally` doesn't throttle. [It keeps track of consumption](https://github.com/mollie/mollie-api-node/pull/283/files#diff-97360564710ef1f5c97fb64333a27c7bbde04935a821651458a7d8877f215b3aR22). The `throttle` function throttles if there is any overconsumption. So while it is true that putting `tally` before or after `yield` can make a big difference in when it was called, the effect is the same.

I think I should maybe document this better, but the idea is _not_ that the throttler throttles the yielding. If 100 payments were downloaded, but `valuesPerMinute` is set to 50, it just yields all of them immediately… but waits a bit before making the _next_ request. Only yielding 50 values even though 100 are available seems a bit childish and unproductive.

Does any of this clear anything up?

src/communication/NetworkClient.ts
```diff
@@ -161,28 +145,76 @@ export default class NetworkClient {
     return embedded[binderName] as R[];
   }
 
-  iterate<R>(relativePath: string, type: string, query: Record<string, any> = {}): HelpfulIterator<R> {
+  iterate<R>(pathname: string, binderName: string, query: Maybe<SearchParameters>, valuesPerMinute = 500): HelpfulIterator<R> {
```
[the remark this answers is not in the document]
Pimm: 2: `query` was optional, because the `iterate` methods in the binders used to kind-of just forward their arguments into `NetworkClient.iterate`. Now they extract `valuesPerMinute` anyway, so there will always be a `query`. (Come to think of it, it doesn't even have to be `Maybe`d any more.)

1: The `500` is not based on anything. The good news is that we can always change these values when we get new insights, it wouldn't cause a breaking change. The bad news is that we have little data to go by.

There are several instances where someone made a bunch of requests to list endpoints (I don't have a number more specific than "a bunch") and caused a spike, making the Mollie API less responsive. That's what we're trying to avoid. I would guess that requesting, say, 2000 values in a minute should be fine.

I actually dislike the "per minute" part as well. A minute feels long. Maybe we should throttle in multiples of 20 seconds instead of minutes, but still advertise it as "per minute" (and divide the `valuesPerMinute` by three)?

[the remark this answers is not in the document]
janpaepke: 2: I see. in that case yes, I agree it shouldn't be `Maybe`

1: Agree on setting a higher default value, although I still think this would be best defined on "hard evidence".
The term `valuesPerMinute` is fine and I don't think it's confusing, even if the throttled chunks come in with less time apart, which I also think its a good idea. 100 values every 10 seconds would still be 600 `valuesPerMinute` and it's a lot more intuitive than 100 `valuesEvery10Seconds`.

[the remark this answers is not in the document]
Pimm: I just changed the default to 5000 values per minute. Additionally, throttling now happens in increments of six seconds instead of minutes. As a result, the throttling now defaults to 500 values per six seconds in practice.

src/communication/NetworkClient.ts
```diff
@@ -161,28 +145,76 @@ export default class NetworkClient {
...
+        // values each.
+        popLimit = () => pageSizes.pop() ?? 128;
       }
-      return new HelpfulIterator<R>(iterate(this.axiosInstance, type, `${relativePath}${stringifyQuery({ ...query, limit })}`));
+      // Set up the throttler. The iterator will wait before making a request to the Mollie API if more values have
+      // been consumed than "allowed" by the valuesPerMinute argument. Note that this iterator will not interrupt the
+      // yielding of values which have already been received from the Mollie API. If a page of 250 values is received
+      // and valuesPerMinute is set to 100, all 250 values received values will be yielded before the (two-minute)
+      // break.
+      const throttler = new Throttler(valuesPerMinute);
```
edorivai, line 177: I may have overlooked this, but are we testing the throttling logic specifically?

edorivai (approved):
Left one comment on testing, but otherwise: LGTM :rocket:

src/communication/NetworkClient.ts
Pimm: We're not testing the timing specifically, no. That would be nice…

Pimm: I added a throttling test. As a nice side-effect, it replaces the previous "multipage" test with a much simpler implementation.

PR #279, opened 2022-07-12
Add `includeWallets` to ListParameters interface

shwao:
This adds the missing `includeWallets` property to the `ListParameters` interface used by `mollie.methods.list`.

PR #285, opened 2022-07-21
Change orderLineCategories in the methods binder to an array

Pimm:
The Mollie API expects `orderLineCategories` in the list methods endpoint to be [a comma-separated list](https://docs.mollie.com/reference/v2/methods-api/list-methods#parameters). However, an array of strings is the more JavaScript-native type here, and the `NetworkClient` will join it anyway.

This is not a breaking change, as `orderLineCategories` was introduced in #279, which has not yet been released.


edorivai (approved):
Lgtm

PR #287, opened 2022-08-01
Fix bug on Node.js < 12

Pimm:
Inspecting objects (payments, customers, subscriptions, etc.) would throw an error on Node.js < 12, because the `inspect.custom` implementation relied on both `Object.entries` and `Object.fromEntries`.


edorivai (approved):

edorivai:
Approved the change, because the change itself looks good, but did you check for other occurrences of `Object.*` functions that are not supported by Node <12?

Pimm:
I just checked. The only other functions we're using are `Object.assign` and `Object.create`, so we should we OK.

We do test on Node.js 6. It just turns out there is no `.toString()` in the test suite. Come to think of it, I might as well add that.

PR #294, opened 2022-09-21
Add retrying functionality

### LINKED ISSUE #257: Investigate retrying on connection errors
I'm not sure what the library currently does if it can't connect to the Mollie servers. We could consider adding retrying functionality for when a request fails early (and we're absolutely certain it didn't reach the Mollie servers) ‒ assuming axios doesn't already do that.

Pimm:
> Note: this PR is currently branched off `pimm/profiles-settlements`. Please merge #293 before this.

Resolves #257.

### Concept

Every requests to which the Mollie API responds with a 5×× status code will be re-attempted until:
 * the Mollie API responds with a different status code, or
 * the attempt limit has been reached, or
 * the request has timed out (as per the timeout set in the Axios instance).

The idea is that if the Mollie API has a brief hiccup, the extra attempts may cause the request to succeed anyway.

For `POST` and `DELETE` requests, an idempotency key is added (in the form of an `Idempotency-Key` header). This ensures the Mollie API can distinguish a single request being re-attempted from two separate similarly-looking requests. In effect, this allows this client to safely re-attempt requests.

We agreed on an attempt limit of **3** and a (fallback) retry delay of **2 seconds**.

We've decided not to retry on network issues, such as no network being available or the request timing out. This client is designed to be run on servers. Unlike on mobile phones with unreliable wireless connections, we expect the network to either work or not work. We don't expect a retry to resolve such network issues.

### Limitations

Because it was really easy to implement, we added support for the [`Retry-After` header](//httpwg.org/specs/rfc9110.html#field.retry-after). I haven't seen the Mollie API actually utilise this header, but it _can_ use it to instruct the client to wait longer or shorter than two seconds.

But… because we can't imagine it ever being useful, we left out support for the HTTP date form of `Retry-After`. If support is ever needed, [there's a library out there to help out](//null.house/open-source/parsehttpdate).

### Note

If we do ever decide to retry on timeouts, we should introduce a per-attempt timeout separate from the global timeout which currently exists. If a developer specifies a `timeout` in their call to `createMollieClient`, retrying after that timeout would be unexpected.


src/communication/makeRetrying.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+    // Determine the delay after which the next attempt is made, preferring a Retry-After header defined in the
+    // response over the default value.
+    const delay = parseRetryAfterHeader(error.response) ?? retryDelay;
+    // Schedule the attempt.
+    return new Promise(resolve => setTimeout(() => resolve(axiosInstance.request(config)), delay));
+  });
+  // Overwrite the post and delete functions in the Axios instance, making them add the header. Note that the Axios
+  // instance still has its original request function, which is not overwritten. Additionally, the Axios instance
+  // itself can be called as a function, which is equivalent to calling the request function. In this client, neither
```
edorivai, line 106: Is this a ticking time bomb? Or do we think this is okay, because worst case a call won't be retried?

edorivai (approved):
LGTM, great work man.

src/communication/makeRetrying.ts
Pimm: A ticking time bomb in the sense that it explodes if someone in the future does `axiosInstance.request` instead of `axiosInstance.post` or `axiosInstance.delete`? Then yes.

The worst case scenario would be bad in that case. It _will_ still retry (because the interceptors are still in effect) but it _won't_ add the idempotency key header. This means any `POST`s or `DELETE`s would be unsafe (as the Mollie API wouldn't be able to distinguish retries from separate requests).

On the bright side: our test suite does specifically check that the header is set.

What are your thoughts after hearing this?

edorivai: I'm not so concerned that consumers of the library will use the axios instance directly. I'm more concerned that someone adds functionality to this library, and uses `axiosInstance.request(...)`. How hard would it be to intercept those calls as well?

Pimm: We're not exposing the Axios instance. (It is technically available as `mollieClient.payments.networkClient.networkClient.axiosInstance`, but using it means ignoring all documentation and type definitions, and is not realistic.)

We could override `axiosInstance.request` instead of `axiosInstance.post` and `axiosInstance.delete`, as those use `.request` internally. However, the Axios instance itself is also callable: `axiosInstance(…)`. We would have to look into overriding the behaviour there as well.

edorivai: Right. Well with that backdrop, my gut feeling is that there is some merit in overriding `.request()` instead of `.post()` and `.delete()`. I can realistically imagine a future contributor using `.request()` instead of `.post()`.

It's unfortunate that it's difficult to override calling the instance, because there's actually no reason that `.request()` is more likely to be used than `axiosInstance(...)`. So we would definitely want some documentation or comments on this.

mollierobbert:
One of our test merchants is asking whether we can tag this commit so they can already try it out, e.g. if we tag it as `prerelease` they can install it via `@mollie/api-client@prerelease`.

mollierobbert:
@Pimm I'm getting the feedback from our test merchant that they would like to configure their own idempotency key as well. Can you make this happen?

This way, they can tie the key to their own back-end. For example if the whole container dies during the request, there is no way of retrying the request if the library only generates its own keys. They would rather generate their own key and provide this to the library explicitly, so they can protect themselves against scenarios that are outside the control of the Mollie library.

Pimm:
> 



> @Pimm I'm getting the feedback from our test merchant that they would like to configure their own idempotency key as well. Can you make this happen?

Thanks for the feedback. Can we schedule a call real quick?

janpaepke:
I think all these ideas are sensible and since they're not mutually exclusive I would suggest to implement them all.

Retrying after certain timeouts seems like a good idea but would be outside the scope of this PR.

Just to clarify: we would still keep the "invisible" feature of supplying our own idempotency key, right? We'd just allow for it to be overwritten either by supplying one or providing a generation callback.

This also means if a consumer both sets up custom key generation and provides a key with the call the latter "wins".

This does not seem counterintuitive but the hierarchy needs to be documented.

src/communication/makeRetrying.ts
```diff
@@ -74,8 +83,13 @@ function parseRetryAfterHeader(response: AxiosResponse): number | undefined {
  * safely re-attempt requests.
  */
 export default function makeRetrying(axiosInstance: AxiosInstance) {
-  // Set the attempt (in the request configuration) of any request.
+  // Intercept all outgoing requests.
   axiosInstance.interceptors.request.use((config: AxiosRequestConfig & Partial<AttemptState>) => {
+    // If the request is a POST or DELETE one and does not yet have the idempotency header, add one now.
+    if (unsafeMethods.has(config.method!) && config.headers?.[idempotencyHeaderName] == undefined) {
+      config.headers = { ...config.headers, ...generateIdempotencyHeader() };
```
janpaepke, line 90:
```suggestion
      Object.assign(config.headers, generateIdempotencyHeader());
```

Pimm: Axios merges the headers passed into `axios.request` with the ones passed to `axios.create`, so the the object passed to the interceptor _should_ be a fresh copy anyway and therefore be safe to mutate.

Is _should_ certain and safe enough for us?

src/communication/makeRetrying.ts
Pimm: We've changed the implementation to use an interceptor instead of overriding any functions, defusing your time bomb.

src/communication/makeRetrying.ts
Pimm: I agree that it reads better. And I might be worrying too much, so let's go for it.

Pimm:
The more I think about idea 2 ‒ the `idempotencyKeyGenerator` ‒ the less I like it.

Here's why:
1. It only adds any value if the engineer wants to derive the idempotency key from the (request) data. Using (a hash of) the internal order ID as the key is something I can envision. But I think this is an uncommon case. I think it's more common that the key is separate from the (request) data, and in that case the generator doesn't help.
2. If the engineer wants to derive some of the idempotency keys from the (request) data, they might still want to use random ones for other endpoints (the default behaviour). That demands some kind of escape hatch to fall back to the default behaviour, thus making it more complicated.
3. If we don't add this functionality, the engineer could work around it:
```javascript
function createPayment(data) {
  return mollieClient.payments.create({
    ...data,
    idempotencyKey: hash(data.metadata.orderID)
  });
}
```

janpaepke:
After discussing this more I tend to agree with this assessment.

It's not that we can't think of any theoretical cases where defining a key generation function  **might** be useful. It's that these cases are very rare and there are valid work-arounds available (see @Pimm's example).

Usually I would say "well if it doesn't hurt, let's just add it". However with this SDK we also have to consider the long term effects of exposing new API. This would have to be supported potentially indefinitely, because one engineer is using it, who could have achieved the same with slightly different code, had he been forced to.

Additionally not adding the generator function makes for a simpler API and less confusion regarding my [point above](https://github.com/mollie/mollie-api-node/pull/294#issuecomment-1327905714).

With those points in mind, I would suggest to leave out idea 2 and only implement 1 and 3.

mollierobbert:
So if I understand correctly, you want to move forward with:

- Auto-generating the idempotency key if none was provided, and automatically use it in the timeout scenario.
- Allowing the user to manually provide a custom key with each API call.
- Exposing the key back to the user in case of an error, so the user can trigger custom retry logic if the auto-retry fails.

All of these make sense to me. 👍

src/communication/makeRetrying.ts
janpaepke: Turns out this doesn't work after all, as `headers` could be `undefined`. 🙄 
I'll fix this in my next pr.

janpaepke:
After finishing Idea 3, I noticed that the `ApiError` wasn't exposed from the lib. Additionally there seems to have been an [outdated interface](https://github.com/mollie/mollie-api-node/blob/master/src/data/global.ts#L147), which was never (externally) exported.
Exporting the error allows for type-safe checking like this:
```typescript
try {...} catch (error {
    if (error instanceof MollieApiError) {
        console.log(error.statusCode /* <- is a known property */);
    }
}
```
Consequently I included exposure of the `ApiError` as `MollieApiError` in the PR: https://github.com/mollie/mollie-api-node/pull/298/commits/aa8f4924b013a67be6bc8a0a6bff6a7f27a6a35c

Pimm:
> I noticed that the `ApiError` wasn't exposed from the lib.

Someone else actually [noticed that just a tad before you did](//github.com/mollie/mollie-api-node/pull/297). But that's good! It just confirms we're on the right track.

> Additionally there seems to have been an [outdated interface](https://github.com/mollie/mollie-api-node/blob/master/src/data/global.ts#L147), which was never (externally) exposed, either.

Good catch!

src/communication/makeRetrying.ts
Pimm: (We've changed this to `Object.assign(config.headers ??= {}, generateIdempotencyHeader())` in [the next PR](https://github.com/mollie/mollie-api-node/pull/298).)

PR #298, opened 2022-11-29
added idempotencyKey prop to ApiError

janpaepke:
For Details see https://github.com/mollie/mollie-api-node/pull/294#issuecomment-1326326017 (Idea 3).

Changes:
- ApiError constructor params are moved into an object
  1. With the addition of `idempotencyKey` the parameter list got unreadably long
  2. The order of the parameters does not matter
- passing `idempotencyKey` from config in static `ApiError.createFromResponse`.
- reverted header assignment (see https://github.com/mollie/mollie-api-node/pull/294#discussion_r1033311534)
- exposing `ApiError` as `MollieApiError` for type-safe checks.
- added tests


src/communication/NetworkClient.ts
```diff
@@ -219,7 +219,7 @@ export default class NetworkClient {
     });
   }
 
-  async patch<R>(pathname: string, data: {}): Promise<R> {
+  async patch<R>(pathname: string, data: NonNullable<unknown>): Promise<R> {
```
Pimm, line 222: This is definitely something we should brainstorm about for a little bit.

The issue is that in TypeScript (at this time), you can't `Exclude` from `unknown` or `any`. `NonNullable<unknown>` is effectively equal to `unknown`, and ‒ surprisingly ‒ `null` is actually an acceptable value for it.

I'm not entirely sure what the most helpful types for `data` in `post` and `patch` and `context` in `delete` would be. I think we're looking for "any object which might have an `idempotencyKey` property" for `post` and `delete`; and "any object" for `patch`. Do you have any thoughts on this?

src/communication/makeRetrying.ts
```diff
@@ -87,7 +87,7 @@ export default function makeRetrying(axiosInstance: AxiosInstance) {
   axiosInstance.interceptors.request.use((config: AxiosRequestConfig & Partial<AttemptState>) => {
     // If the request is a POST or DELETE one and does not yet have the idempotency header, add one now.
     if (unsafeMethods.has(config.method!) && config.headers?.[idempotencyHeaderName] == undefined) {
-      Object.assign(config.headers, generateIdempotencyHeader());
+      config.headers = {...config.headers, ...generateIdempotencyHeader()}
```
Pimm, line 90: One alternative would be:
```javascript
Object.assign(config.headers ??= {}, generateIdempotencyHeader());
```
Too far-fetched?

Pimm: Actually, `headers` can actually never be `undefined` as long as default headers are provided, which we do in `NetworkClient`. [Axios' source code](//github.com/axios/axios/blob/412d3bd6078433dda4a7eb4d86e021dbc57fdeb2/lib/core/mergeConfig.js) confirms this.

Does that change our mind?

src/errors/ApiError.ts
```diff
@@ -1,13 +1,31 @@
...
+  statusCode?: number;
+  idempotencyKey?: string;
+  title?: string;
+  links?: MollieApiErrorLinks;
+};
+
 export default class ApiError extends Error {
-  public constructor(message: string, protected title?: string, public readonly statusCode?: number, public readonly field?: string, protected links?: MollieApiErrorLinks) {
+  // Set the name to ApiError.
+  public readonly name: string = 'ApiError';
```
Pimm, line 18: `Error` actually comes with a `name` property [out of the box](//tc39.es/ecma262/#sec-error.prototype.name). I think we should use that instead of adding our own.

janpaepke: discussed:
+ easier to set `Symbol.toStringTag`
- subtle usage differences

=> Decision to keep it like this.

src/communication/makeRetrying.ts
janpaepke: discussed, decision to keep this version as it's the most future proof.
```Object.assign(config.headers ??= {}, generateIdempotencyHeader());```

src/communication/NetworkClient.ts
Pimm: It turns out that [since TypeScript 4.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-8.html), `unknown` is "close in spirit" to `{} | undefined | null`. This means `NonNullable<unknown>` is ‒ since 4.8 ‒ equal to `{}`.

However, we've decided to go with `Record<string, unknown>`. Although Axios can send a broader variety of data, this library will only send (string-keyed) objects.

Pimm:
We still need this, GitHub. Thanks.

Pimm (approved):
I like how the error handling code is not unnecessarily interwoven with `makeRetrying`.

PR #311, opened 2023-03-08
Update documentation

Pimm:
Copied a fresh copy of the documentation from [docs.mollie.com](https://docs.mollie.com).


dsempel (approved):
👍

PR #314, opened 2023-03-10
Remove snake_case client properties

Pimm:
The snake case client properties (such as `customers_payments`) were renamed to dromedary case (such as `customerPayments`) [a while ago](https://github.com/mollie/mollie-api-node/pull/223). The first version to include this rename is 3.6.0, released on 10 February 2022.

This commit removes the snake case aliases.

The change is breaking. The next release will be 4.0.0.


dsempel (approved):
Lgtm

PR #315, opened 2023-03-10
Remove endpoint aliases

Pimm:
These endpoint aliases (such as `client.payments.delete` for `client.payments.cancel`) were deprecated in #234. The first version to include this deprecation is 3.6.0, released on 10 February 2022.

This PR removes the aliases.

Rationale:
* I suspect the aliases were introduced with the idea that "someone might misremember and accidentally type `delete` instead of `cancel`". We now live in the era of TypeScript, we no longer rely on memory; we rely on code completion. When one relies on code completion, aliases _introduce_ confusion: "`delete` sounds like what I need, but so does `cancel`… what's the difference between the two? When should I use which? Why am I using `delete` while this snippet on StackOverflow uses `cancel`?"
* The `all` aliases have always been deceptive. Instead of returning _all_ objects in an array, they were aliases for `page` and thus returned _the first page_ in an array.
* Because a binder now has either `page` or `list`, you can see whether the result will be paginated or not. 

The change is breaking. The next release will be 4.0.0.

### DIFF
--- a/examples/chargebacks/list.ts
+++ b/examples/chargebacks/list.ts
@@ -7,7 +7,7 @@ const mollieClient = createMollieClient({ apiKey: 'test_dHar4XY7LxsDOtmnkVtjNVWX
 
 (async () => {
   try {
-    const chargebacks: List<Chargeback> = await mollieClient.chargebacks.all();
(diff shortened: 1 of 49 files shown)


edorivai (approved):
Lgtm, nice cleanup

PR #317, opened 2023-04-21
fix small type issue

janpaepke:
The createdAt property of subscriptions is actually not optional.
(see [docs](https://docs.mollie.com/reference/v2/subscriptions-api/get-subscription?path=createdAt#response))


Pimm (approved):
Perfect!

([A PR](//github.com/mollie/api-documentation/pull/940) for `canceledAt` was created.)

Issue #291, opened 2022-08-29
Respect pagination in helpers


Pimm:
The following helper functions return a (plain) array, although the underlying endpoint is [paginated](https://docs.mollie.com/overview/pagination):
 * `customer.getMandates`,
 * `order.getRefunds`,
 * `payment.getChargebacks`,
 * `payment.getRefunds`,
 * `profile.getChargebacks`,
 * `profile.getPayments`,
 * `profile.getRefunds`, and
 * `subscription.getPayments`.

This means the lists returned by these functions may not be exhaustive, which may surprise developers.

For some functions ‒ such as `customer.getMandates` or `order.getRefunds` ‒ this is realistically not an issue. It is unlikely that a single customers has enough mandates for the API to start paginating them, or for a single order to have enough refunds for the API to start paginating them. However, for `profile.getPayments`, the scenario is more likely.

I see three possible solutions:

### Introduce `nextPage`

We could (very easily!) add the `nextPage` function to the returned lists. This will allow developers to do:
```javascript
var payments = await profile.getPayments();
payments = await payments.nextPage();
```
However, developers may not notice the existence of the `nextPage` function. And if they don't, we're back to square one. We could _additionally_ rename `profile.getPayments` to `profile.pagePayments` to convey the paginated nature.

It would be an easy solution, but a breaking change. And in my opinion, not a pretty API.

### Make the arrays exhaustive

We could repeatedly call the endpoint until the list is depleted. The returned arrays would always be exhaustive. This would not cause a breaking change. However, the helper functions would become potentially expensive and might put a lot of stress on the Mollie API.

### Have the helper functions return iterators

Developers would have to do:
```javascript
for await (let payment of profile.getPayments()) {
  […]
}
```
instead of:
```javascript
for (let payment of await profile.getPayments()) {
  […]
}
```

It would be a breaking change as well.

The only other downside to this solution I can see, is that ‒ as some other helper functions rely on endpoints which are not paginated ‒ the API would be inconsistent. `profile.getPayments` would return an iterator, while `profile.getMethods` would return an array. (We could consider returning iterators from _all_ list-y helper functions.)

edorivai:
My knee-jerk reaction would be to consistently have the helper function return iterators. This should feel familiar for consumers that have been using the `.iterate()` methods. It's also the best fit with the current API IMO.

[...]

PR #322, opened 2023-06-19
List API changes

Pimm:
### Plain array return types

Changed the return type of `methods.list`, `orderShipments.list`, and `permissions.list` to (plain) arrays.

Previously, the object returned by those methods had optional properties, including `nextPage`:
```typescript
const methods = await methods.list();
await methods.nextPage?.();
```
As the underlying endpoints are not paginated, `nextPage` was always `undefined` anyway. Removing the `nextPage` property communicates this better.

### Iterator return types

Changed the return type of several helper functions to iterators, including `customer.getPayments`.

This forces users to handle the scenario in which the items do not fit on a single page, e.g. one customer has more than 250 payments. See #291.

### Renames

Renamed types and methods to solidify the distinction between (exhaustive) lists and pages. This does not affect the public-facing API (unless you count the exported TypeScript types).

---

Some changes are breaking. The next release will be 4.0.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "lint": "yarn lint:eslint:fix && yarn lint:prettier"
   },
   "dependencies": {
-    "axios": "^0.27.2"
+    "axios": "^0.27.2",
+    "ruply": "^1.0.1"
   },
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.18.6",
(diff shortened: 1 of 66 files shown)


edorivai:
LGTM, only note I have is that I would make this part of a major version upgrade since it will break TS builds.

edorivai (approved):

Pimm:
> […] I would make this part of a major version upgrade since it will break TS builds.

Yes! I changed the description of the PR accordingly.

PR #323, opened 2023-06-20
Remove withParent

Pimm:
This library relies heavily on the network, which makes it asynchronous throughout. Asynchronicity doesn't mix well with stateful operations such as `withParent`.

While this code will always work:
```typescript
const payment = await mollieClient.customerPayments.withParent({ id: 'cst_pzhEvnttJ2' })
                                                   .create([…]);
```
This code may be surprising:
```typescript
mollieClient.customerPayments.withParent({ id: 'cst_pzhEvnttJ2' });
// […]
const payment = mollieClient.customerPayments.create([…]);
```
If `withParent` is called after that first line executes but before the second one does, it will overwrite the customer. If the `await` keyword appears between those two lines, that `withParent` call could be _anywhere_.

The stateless variant does not have this issue:
```typescript
const payment = await mollieClient.customerPayments.create({ customerId: 'cst_pzhEvnttJ2', … });
```

This change is breaking. The next release will be 4.0.0. `withParent` was deprecated in 2018.


dsempel (approved):
This makes sense to me 👍

PR #330, opened 2023-09-08
Added captures as embedded objects to payment

JoostMollie:


Pimm:
Thanks, @JoostMollie!

We should also change `PaymentHelper.getCaptures`. It should use the embedded captures if present to prevent unnecessary network access. (I'm OK doing that.)

However, [the documentation currently does not mention embedded captures](//docs.mollie.com/reference/v2/payments-api/get-payment?path=cancelUrl#embedding-of-related-resources). Is this a new thing? Is it stable?

JoostMollie:
> Thanks, @JoostMollie!
> 
> We should also change `PaymentHelper.getCaptures`. It should use the embedded captures if present to prevent unnecessary network access. (I'm OK doing that.)
> 
> However, [the documentation currently does not mention embedded captures](//docs.mollie.com/reference/v2/payments-api/get-payment?path=cancelUrl#embedding-of-related-resources). Is this a new thing? Is it stable?

Yes the embedding of captures in a payment is stable. We will update the public docs soon. Our private/beta docs are already updated.

Pimm (approved):
Alright, then!

Issue #349, opened 2024-06-19
TypeError: Missing parameter "apiKey" or "accessToken".


vrxj81:
I'm trying to us the Mollie NodeJS API in a NestJS application, but when I use the createMollieClient function, I get the error **TypeError: Missing parameter "apiKey" or "accessToken".** while I'm providing the apiKey options.
See code below:
```import { Inject, Module } from '@nestjs/common';
import { PaymentsService } from './payments.service';
import {
  ConfigurableModuleClass,
  OPTIONS_TYPE,
  PAYMENT_MODULE_OPTIONS,
} from './backend-payments-util.module.definition';
import { MolliePaymentsProvider } from './providers/mollie.payments.provider';
import createMollieClient, { MollieClient, MollieOptions } from '@mollie/api-client';

@Module({})
export class BackendPaymentsUtilModule extends ConfigurableModuleClass {
  constructor(
    @Inject(PAYMENT_MODULE_OPTIONS) private options: string | symbol,
  ) {
    super();
  }

  static register(options: typeof OPTIONS_TYPE) {
    if (options.paymentProvider === 'mollie') {
      **const mollieClient: MollieClient = createMollieClient({
        apiKey: process.env['MOLLIE_API_KEY'] || '',
      } as MollieOptions);**
      return {
        ...super.register(options),
        providers: [
          {
            provide: PAYMENT_MODULE_OPTIONS,
            useValue: options,
          },
          {
            provide: PaymentsService,
            useFactory: () => new MolliePaymentsProvider(mollieClient),
          },
        ],
      };
    } else {
      throw new Error('Payment provider not supported');
    }
  }
}
```

I also tried the createMollieClient available like so:
`import { createMollieClient,  MollieClient, MollieOptions } from '@mollie/api-client';`
But same issue.

Pimm:
Thanks for reaching out, Johan.

The current version of the client performs a [falsy](https://developer.mozilla.org/docs/Glossary/Falsy) check on `apiKey` and `accessToken`. It will throw the "missing parameter" error for any falsy value, including an empty string. Could you check the value of `process.env['MOLLIE_API_KEY']` (using `console.log`, for instance)?

Pimm:
I'm closing this issue, as I suspect you accidentally passing an empty string to `apiKey` is what is causing your issue.

If I'm wrong, please comment and I'll reopen.

Hope your application turns out nicely!

PR #350, opened 2024-07-05
More helpful error in createMollieClient for missing credentials

Pimm:
Previously, not setting `apiKey` or `accessToken` to an appropriate value would result in a generic error. After this PR, an accidental empty string (`createMollieClient({ apiKey: '' })`) gives a distinct error from no value at all (`createmollieClient({})`).

I suspect this is what happened in #349, and I hope this will prevent some headaches in the future.

### DIFF
--- a/src/createMollieClient.ts
+++ b/src/createMollieClient.ts
@@ -4,6 +4,7 @@ import caCertificates from './cacert.pem';
 import NetworkClient from './communication/NetworkClient';
 import TransformingNetworkClient, { Transformers } from './communication/TransformingNetworkClient';
 import type Options from './Options';
+import { run } from 'ruply';
 
 // Transformers
 import { transform as transformPayment } from './data/payments/Payment';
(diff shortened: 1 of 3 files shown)


src/createMollieClient.ts
```diff
@@ -58,6 +59,25 @@ import SettlementsBinder from './binders/settlements/SettlementsBinder';
...
+  ]) as Map<any, string>,
+  descriptions =>
+    function describeFalsyOption<O extends Record<string, unknown>>(object: O, key: keyof O & string) {
+      if (key in object == false) {
+        return null;
+      }
+      return `Parameter "${key}" is ${descriptions.get(object[key]) ?? object[key]}.`;
+    },
+);
+
```
dsempel, line 80: It might be a good idea to extract the credentials check to a separate file. What do you think?

dsempel (approved):
This is a good idea! I added a suggestion.

src/createMollieClient.ts
Pimm: Yes, that's sensible.

PR #351, opened 2024-07-09
Unconditionally run all tests

Pimm:
Previously, some tests were skipped on older Node.js versions because they rely on Nock, which is unsupported by Node.js < 8.0.0.

We are currently figuring out which Node.js versions 4.0.0 of the client will support, but we plan on raising it to at least 8.0.0. We can thus safely run all tests unconditionally.


dsempel (approved):
LGTM

PR #352, opened 2024-07-09
Remove getEntries polyfill (for Object.entries)

Pimm:
[Object.entries](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) is unsupported by Node.js < 7.0.0. We are currently figuring out which Node.js versions 4.0.0 of the client will support, but we plan on raising it to at least 8.0.0. This eliminates the need for a Object.entries polyfill.


dsempel (approved):
LGTM

Issue #319, opened 2023-05-25
Consider introducing runtime backwards compatibility


Pimm:
We are currently introducing [breaking](//github.com/mollie/mollie-api-node/pull/314) [changes](//github.com/mollie/mollie-api-node/pull/315) for the 4.0.0 release, and there are more to come.

These breaking changes:
 * reduce cause for confusion,
 * improve discoverability,
 * improve consistency with the rest of the JavaScript ecosystem, and
 * improve consistency with the PHP version of this client.

These breaking changes will cause compile-time errors for TypeScript users (good, as it facilitates the migration); and runtime errors for JavaScript users (frustrating).

We could consider implementing runtime backwards compatibility features ‒ that is, features which don't affect the exported type declarations. TypeScript users would still get compile-time errors; but JavaScript users _might_ just get a working version without migration.

Obviously, backwards compatibility cannot exist forever. It would be a temporary measure.

janpaepke:
The danger you are referring to here is, that in contrast to TS changes, JS changes might go unnoticed when upgrading.
Adding runtime backwards compatibility (which should produce console warnings) might be a good intermediate step.

However if there are features, where this is not possible (for example if it conflicts with TS changes) I vote strongly against this. The reason being that it gives the user a false sense of security.

An obvious step is to prepare a comprehensive migration guide.
If we expect many users to be affected, we _might_ also consider preparing migration scripts.

In short, if we add backwards compatibility features, it should cover everything. If it doesn't, we should make a clean break.

My gut is leaning towards the latter - this is what major versions are for and we can expect devs to understand that.

Pimm:
> The danger you are referring to here is, that in contrast to TS changes, JS changes might go unnoticed when upgrading.

Exactly. My main concern is that ‒ because we are renaming a lot of things but not _everything_ ‒ a quick and careless test from a user (developer) may give the false impression that their code works with 4.×.× without any changes.

We always write [a comprehensive migration guide](//github.com/mollie/mollie-api-node/blob/master/MIGRATION.md).

> In short, if we add backwards compatibility features, it should cover everything. If it doesn't, we should make a clean break.

Introducing a `customers_payments` → `customerPayments` runtime backwards compatibility hack is trivial. A similar hack for `withParent` would not be. I want to finally get rid of `withParent`: it was poor API design to begin with, and removing it will clean up the codebase significantly. `withParent` runtime backwards compatibility would deprive us of that cleanup.

Pimm:
[...]

PR #353, opened 2024-07-09
Re-introduce runtime aliases

Pimm:
We [removed snake_case aliases for client properties](//github.com/mollie/mollie-api-node/pull/314) and [aliases for endpoints](//github.com/mollie/mollie-api-node/pull/315). This PR re-introduces them, this time unenumerable and non-existent in the TypeScript types.

This will not do much for TypeScript users, they will still get a compile-time error. This merely affects JavaScript users who are still using deprecated aliases. See #319 for more discussion on this.


dsempel (approved):
LGTM

PR #355, opened 2024-08-10
For methods -> pricing changed PricingMethod to array

TimmyGuy:
Following https://docs.mollie.com/reference/v2/methods-api/get-method?path=pricing#response the response is an array of MethodPricing


janpaepke:
This is correct and incontroversial. I will see if we can add some tests and we can ship it with 4.0.0

janpaepke:
closed and replaced by https://github.com/mollie/mollie-api-node/pull/374

Issue #346, opened 2024-05-06
Axios Dependency


marxys:
Hello I saw that mollie-api-node run with an old version of Axios.
Here is my npm audit logs : 

```
Axios Cross-Site Request Forgery Vulnerability - https://github.com/advisories/GHSA-wf5p-g6vw-rhxx
No fix available
node_modules/@mollie/api-client/node_modules/axios
  @mollie/api-client  <=3.7.0
  Depends on vulnerable versions of axios
  node_modules/@mollie/api-client
```
I solve the issue on my side installing this : 
`npm install npm-force-resolutions --save-dev`

Then add this on the package.json : 

```
"resolutions": {
  "axios": "1.6.8"
}
"scripts": {
  // add to existing scripts
  "preinstall": "npx npm-force-resolutions"
}
```
Then `npm run preinstall` to upgrade.

vdhpieter:
@maria-swierblewska @Pimm any timeline on solving this?

n2k3:
Yeah, would love to see this updated as well.

As per: https://github.com/advisories/GHSA-wf5p-g6vw-rhxx the minimum fixed version would be 0.28.

To patch this in your own repo for now (when using `pnpm`), add this to your `package.json`, and run `pnpm i`
```json
  "pnpm": {
    "overrides": {
      "axios": "^0.28.0"
    }
  }
```

Pimm:
Thanks for opening this issue.

The vulnerability in question can potentially reveal a secret stored in a cookie to untrusted servers. This client is not designed to run in a browser, so cookies are not a concern. Furthermore, the client connects to the Mollie server exclusively. The vulnerability therefore does not affect this client or its users, which is why no emergency release was made in response to it.

With that said, we plan to release a new version no later than next week.

dannybunschoten:
Any news on this yet?

Pimm:
We are working on a version of the client which uses [fetch](https://developer.mozilla.org/docs/Web/API/fetch) internally, swapping out Axios altogether. It hasn't been thoroughly tested, but it seems we will go with this approach in 4.0.0.

Until then ‒ provided that you're on Node.js 14+ ‒ you can use the beta version:
```
npm install @mollie/api-client@beta
```

PR #358, opened 2024-09-06
Replace axios with (node-)fetch

janpaepke:


package.json
```diff
@@ -51,7 +51,7 @@
     "@mollie/eslint-config-typescript": "^1.6.5",
     "@types/jest": "^29.5.11",
     "@types/node": "^18.14.6",
-    "axios-mock-adapter": "1.19.0",
+    "@types/node-fetch": "^2.6.11",
```
Pimm, line 54: The `@types` and the regular dependency are out of sync.

janpaepke: It's the same major and at the time of writing there is no newer types package for node-fetch.
This should be fine.

src/Options.ts
```diff
@@ -29,7 +27,7 @@ type Options = Xor<
    * The URL of the root of the Mollie API. Default: `'https://api.mollie.com:443/v2/'`.
    */
   apiEndpoint?: string;
-} & Pick<AxiosRequestConfig, 'adapter' | 'proxy' | 'socketPath' | 'timeout'>;
+};
```
Pimm, line 30: We should document that this is a breaking change.

Pimm:
Fixes https://github.com/mollie/mollie-api-node/issues/346.

fjbender (approved):

src/communication/NetworkClient.ts
```diff
@@ -69,21 +68,37 @@ const throwApiError = run(
...
 interface Data {}
 interface Context {}
 
 /**
- * This class is essentially a wrapper around axios. It simplifies communication with the Mollie API over the network.
+ * This class is essentially a wrapper around fetch. It simplifies communication with the Mollie API over the network.
  */
 export default class NetworkClient {
-  protected readonly axiosInstance: AxiosInstance;
+  protected readonly request: (pathname: string, options?: RequestInit) => Promise<ResponseWithIdempotencyKey>;
```
Pimm, line 101: Could we add one or two sentences to explain what this function does, specifically how it differs from (vanilla) `fetch`?

src/communication/NetworkClient.ts
```diff
@@ -92,71 +107,67 @@ export default class NetworkClient {
...
-      return true;
-    }
-    return response.data;
+    const { idempotencyKey, ...body } = data;
+    const config: RequestInit = {
+      method: 'POST',
+      headers: idempotencyKey ? { [idempotencyHeaderName]: idempotencyKey } : undefined,
+      body: JSON.stringify(body),
+    };
+    return this.request(buildUrl(pathname, query), config).catch(throwApiError).then(processFetchResponse);
```
Pimm, line 148: The `.catch(throwApiError).then(processFetchResponse)` seems repetitive. Can we change the abstraction somehow?

src/communication/NetworkClient.ts
```diff
@@ -69,21 +68,37 @@ const throwApiError = run(
...
+async function processFetchResponse(response: ResponseWithIdempotencyKey) {
+  // Request was successful, but no content was returned.
+  if (response.status == 204) return true;
+  // Request was successful and content was returned.
+  const body = await response.json();
+  if (Math.floor(response.status / 100) == 2) {
+    return body;
+  }
+  // Request was not successful, but the response body contains an error message.
+  if (body) {
```
Pimm, line 87: Can we replace this truth-y check with something else? Are we checking for `undefined` here? Or null-ish?

src/communication/makeRetrying.ts
```diff
@@ -42,83 +31,69 @@ function generateIdempotencyHeader() {
...
   }
   // Convert the header value from seconds to milliseconds.
   return retryAfter * 1e3;
 }
 
+export type ResponseWithIdempotencyKey = Response & { idempotencyKey: string | undefined };
+
 /**
- * Enhances the passed Axios instance, making it attempt requests multiple times in some scenarios.
+ * Wrapper around fetch, making it attempt requests multiple times in some scenarios.
```
Pimm, line 56: Could we change this documentation to start with a verb?

tests/unit/resources/__snapshots__/lists.test.ts.snap
```diff
@@ -64,68 +64,3 @@ exports[`lists .list() should retrieve the next page 1`] = `
   },
 ]
 `;
-
-exports[`lists .list() should retrieve the next page 2`] = `
```
Pimm, line 68: Could we double-check that this is intentional?

janpaepke: Yes, this is indeed intentional, as there were previously two tests doing the same thing and they have been refactored to be one single test, rendering one of the snapshots obsolete.

Pimm:
### A note on Node.js version support.

One could make the argument that periodically increasing the Node.js version this client requires is a good thing, because it motivates users to upgrade their Node.js version to one which is still in the maintenance window. I personally suspect and fear the true effect would instead be that users stop updating this client altogether. I therefore continue to believe the best strategy is supporting all Node.js versions which are still being maintained, as well as any older version we can reasonably support – including through polyfills.

The current oldest Node.js version which is still being maintained is 18, which will be maintained until 30 April 2025. However, through [node-fetch](https://github.com/node-fetch/node-fetch), we can easily support 14 and up.

In fact, node-fetch 2.× supports Node.js all the way down to 6. However, Node.js versions under 14 are challenging to test, as Jest and Nock have version requirements of their own.

We decided to raise the Node.js version requirement to 14, while _secretly_ supporting 8. We've tested the client including this PR on Node.js 8, and it does work, but the test suite itself needs to be modified to make the tests themselves compatible.

PR #362, opened 2024-09-10
Remove toPlainObject function

Pimm:
This PR removes the `toPlainObject` function:
```typescript
const plainPayment = payment.toPlainObject();
```
This function was originally implemented using [Lodash' toPlainObject](https://lodash.info/doc/toPlainObject). I have personally never understood the use case of this function. I suspect this function has very few uses, if any. If my suspicion is wrong, I'll happily re-introduce it.


fjbender (approved):

inglkruiz:
@Pimm Hi, actually in my case this function was extremely useful as I use it to store for instance Payment object in our DB. Later we use it the Object for deriving our own app DB model because we integrate multiple payment providers, Mollie in Europe and Stripe in the US, at the moment.

janpaepke:
Thank you for your feedback, @inglkruiz.

Could you share a little bit more about your implementation details?
Specifically why a simple destructuring would not help in your usecase?
i.e. `storeInDb({...payment})` ?

PR #367, opened 2024-09-10
update metadata type from `any` to `unknown`

janpaepke:
The general direction for ts is to use unknown over any here to force better type validation.

Also see:
https://github.com/microsoft/TypeScript/issues/41016


Pimm:
This is a breaking change. I'll be sure to add to the migration guide that you can `as any` to opt-into type errors.

Pimm (approved):

Issue #264, opened 2022-02-15
Please add PaymentLink to types.d.ts


JohanManders:
Because I have to switch from the Payment API to the Payment Link API, I came across this little annoyance.

For importing Payment I can use:

```ts
import { Payment as MolliePayment } from '@mollie/api-client'
```

So when I switched to Payment Link, I thought I could use:

```ts
import { PaymentLink as MolliePaymentLink } from '@mollie/api-client'
```

But this does not work, because the file ```@mollie\api-client\dist\types\types.d.ts``` does not have this line:

```ts
export { default as PaymentLink } from './data/paymentLink/PaymentLink';
```

Could you please add it, this would make PaymentLink work similar like Payment.

Thanks!

PS It would be great if I could use ```expiresAt``` with the regular Payment API, so that I did not have to switch to the Link version of the API.

Pimm:
Thanks Johan,

I'll export `PaymentLink`!

Could you elaborate on how you would use `expiresAt` for payments?

JohanManders:
Hi Pimm,

I have a case where we want to sell some tickets between ```14:00``` and ```15:00``` and a raffle will be held at ```15:30```, 30 minutes after the last moment someone can buy a ticket.

But, payment methods have different expire times. I am sure that no iDeal payment will come through after ```15:15```, because the expire time is 15 minutes, but for example Credit card (30 minutes) or Paypal (3 hours) can come through (way) after the raffle is done.

We have to inform these people that they paid for something that they did not participate in the raffle and have to pay them back. This is extra hassle...

Payment Link API has a ```expiresAt``` field that helps a bit. The payment link is not valid after this set datetime, but people could open the link, get some beer while leaving their browser tab open and still pay after an hour if they use Paypal for instance.

A solution for the Payment API would be to loop through all pending orders at ```15:15``` and cancel them all, but yeah... it is a bit of a hacky solution in my opinion.

So, I would love to 'say' to Mollie that a payment cannot be done after a certain datetime. That the payment ```expiresAt```, even if someone leaves his browser tab open at a certain Mollie payment page/ link.

janpaepke:
@Pimm please advise how we plan to continue here.

Pimm:
The question about `expiresAt` has been flying under the radar, because it's buried in this ticket about `PaymentLink`. It would be great if we can get closure from the Mollie API team about this.

fjbender:
[...]

Issue #284, opened 2022-07-21
Use type instead of enum


shwao:
Some things that could be a TypeScript `type` are `enum` which is quite developer unfriendly. Code completion and many checks just don't work. 

# The Problem

For example if you have a `Payment` and want to check its status, the IDE (VS Code in my case) can't give you anything to autocomplete. The type check works though.

Here you should have the statuses as suggestions:
![Screenshot 2022-07-21 at 14 48 58](https://user-images.githubusercontent.com/63549997/180217153-87a9e80d-0ac2-427c-ae2b-1d91474b0107.png)

This is the case for every `enum` including `Locale`, `PaymentMethod` etc.

While the missing suggestions are basically just an inconvenience (and missing out on fully utilizing TypeScript and the IDE) using the `embed` option of the `mollieClient.payments.get` method just does not work without using `any` because of the `enum`:

![Screenshot 2022-07-21 at 14 51 51](https://user-images.githubusercontent.com/63549997/180217737-fb9aa4da-368f-4839-b1c0-a9776d810afe.png)

In contrast to the `include` option which uses a `type`:

![Screenshot 2022-07-21 at 14 55 10](https://user-images.githubusercontent.com/63549997/180218449-94ff7c40-4bae-480c-a068-bc8dd0328aea.png)

Changing the `PaymentEmbed` from an `enum` to `type` fixes this:

![Screenshot 2022-07-21 at 15 01 49](https://user-images.githubusercontent.com/63549997/180219850-bc8695bf-8740-49c8-8af4-c682de96e7c2.png)
![Screenshot 2022-07-21 at 15 03 16](https://user-images.githubusercontent.com/63549997/180220135-a84160e9-823a-4124-84a7-4862dc577bc9.png)

# Suggested Changes

Looking at the source code it seems like this change could be easy.

Lets use the `PaymentStatus` as an example.

The `PaymentStatus` `enum` from `src/data/payments/data.ts:825` is used like an Object in `src/data/payments/PaymentHelper.ts` to check against the payment status.

I would suggest to change `src/data/payments/data.ts` from this:

```ts
export enum PaymentStatus {
  open = 'open',
  canceled = 'canceled',
  pending = 'pending',
  authorized = 'authorized',
  expired = 'expired',
  failed = 'failed',
  paid = 'paid',
}
```

To this:

```ts
export const PAYMENT_STATUS = {
  open: 'open',
  canceled: 'canceled',
  pending: 'pending',
[...]

PR #370, opened 2024-09-11
Payments: add missing include type

janpaepke:


Pimm (approved):
Turning `PaymentInclude` into an enum is consistent with most of our existing API, but a discussion is to be had [about using enums vs string constants](https://github.com/mollie/mollie-api-node/issues/284).

Pimm:
Fixes #264.

janpaepke:
> Fixes #264.

How does this fix #264 ? That one is about `PaymentLink`, this one is about `PaymentInclude`.

src/data/payments/data.ts
```diff
@@ -862,7 +862,7 @@ export enum PaymentStatus {
 
 export enum PaymentInclude {
   qrCode = 'details.qrCode',
-  remainderDEtails = 'details.remainderDetails',
```
janpaepke: woops 😅

Issue #356, opened 2024-08-15
Expose MandateDetailsCreditCard type


jp1987:
When using TypeScript, I can't seem to import MandateDetailsCreditCard as a type through `@mollie/api-client`, but only like so:

`import type { MandateDetailsCreditCard } from '@mollie/api-client/dist/types/src/data/customers/mandates/data';`

janpaepke:
@Pimm: We mostly export enums from the data, but I don't think it would hurt to expose those types as well. Hence the pr.

Pimm:
@janpaepke I think it _could_ hurt, as `MandateDetailsCreditCard` is intended to be an internal type. Exporting it would make it part of the public API, which means we'll have to maintain it.

@jp1987 Could you elaborate on how you are currently using `MandateDetailsCreditCard`?

jp1987:
> @jp1987 Could you elaborate on how you are currently using `MandateDetailsCreditCard`?

in a custom mapper:

```typescript
export const mapCreditCardMandates = (client: IClient, mandate: Mandate) => {
  const mandateDetailsCreditCard = mandate.details as MandateDetailsCreditCard;
  return {
    active: client.mollieActiveMandateId === mandate.id,
    cardHolder: mandateDetailsCreditCard.cardHolder,
    cardLabel: mandateDetailsCreditCard.cardLabel,
    cardNumber: mandateDetailsCreditCard.cardNumber,
    mandateId: mandate.id,
    status: mandate.status,
  };
};
```

`MandataDetails` is exported like this: `export type MandateDetails = MandateDetailsCreditCard | MandateDetailsDirectDebit;` so we need to force the type.

Pimm:
Thanks for elaborating, Jesper. I think this warrants a conversation about the types we (don't) want to expose.

A discriminated union might actually help in this specific case:
```typescript
if (mandate.method == 'creditcard') {
  mandate.details.cardHolder // ← guaranteed to exist inside this if-block.
}
```

janpaepke:
Not only would this help, but it would make your code more stable.

@Pimm do you want to have this conversation here or open up a separate issue for this?

PR #372, opened 2024-09-11
Mandates: export missing mandate types from package

janpaepke:
This fixes https://github.com/mollie/mollie-api-node/issues/356


Pimm (changes requested):
These types are intended to be internal.

I would like to better understand how and why they are to be used before we expose them, making them part of the public API.

janpaepke:
Closing this PR to keep things neat.
We can reopen, should the discussion in https://github.com/mollie/mollie-api-node/pull/372 conclude to warrant this.

Issue #316, opened 2023-03-10
Property "issuers" doesn't exist on type "Method"


stefan-at-junda:
I'm using your module with TypeScript, and when I include the issuers while fetching the payment methods and loop over them, TypeScript says that "issuers" doesn't exist on the "Method" type. It seems like this property is missing there.

janpaepke:
To confirm, did you use `include: MethodInclude.issuers` in your request?

PR #374, opened 2024-09-11
Methods: fix types for pricing and issuer

janpaepke:
When using the method's `include` option, we expect the result to include a `pricing` or `issuers` property, which is an array of objects.

see https://docs.mollie.com/reference/get-method

This replaces https://github.com/mollie/mollie-api-node/pull/355 and resolves https://github.com/mollie/mollie-api-node/issues/316


Pimm (approved):
LGTM.

PR #376, opened 2024-09-17
Update payment methods

### LINKED ISSUE #357: Please update the payment methods
Some payment methods have been removed and some payment methods have been added to mollie, maybe its time to update it and add all new payment methods and remove old ones.(Giropay and sofort is no longer available.)

janpaepke:
Fixes #357 

The reference list is taken from the [API docs for the get method](https://docs.mollie.com/reference/get-method).

1. **Added the following new payment methods to `PaymentMethod`**:
   - alma
   - bacs
   - bancomatpay
   - billie
   - blik
   - riverty
   - satispay
   - trustly
   - twint

2. **Moved outdated payment methods to `HistoricPaymentMethod`**:
   - giropay
   - sofort

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,12 @@
 
 Please see [the migration guide](MIGRATION.md) for guidance about updating to a newer major version.
 
-### v4.0.0 - 2024-09-16
+### v4.0.0 - 2024-09-19
   - Replace Axios dependency in favour of [fetch](https://developer.mozilla.org/docs/Web/API/fetch) ([#358](https://github.com/mollie/mollie-api-node/pull/358))
   - Add `cancelUrl` and `getDashboardUrl` to payments and orders ([#327](https://github.com/mollie/mollie-api-node/pull/327)/[#373](https://github.com/mollie/mollie-api-node/pull/373))
   - Add `status` and `issuers` to methods and update `pricing` ([#335](https://github.com/mollie/mollie-api-node/pull/335)/[#374](https://github.com/mollie/mollie-api-node/pull/374))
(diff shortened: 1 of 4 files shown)


fjbender (approved):

Pimm:
@janpaepke mentioned that this PR Introduces a breaking change, but I'm unsure whether I agree.

The following code compiles before this PR, but not after:
```typescript
mollieClient.payments.create({
  method: PaymentMethod.giropay,
  …
});
```
Existing code no longer compiler is the tell-tale of a breaking change.

_However_, while the code above does _compile_ before this PR, it doesn't work anyway, [as Giropay has been deprecated](https://help.mollie.com/hc/en-us/articles/19745480480786-Giropay-Depreciation-FAQ).

I am currently on the fence whether this change ‒ and similar future changes ‒ should be considered "breaking" and thus warrant a major version bump.

janpaepke:
It was not marked as deprecated in the code and thus a consumer might likely be unaware.
Imho in this particular example it's pretty clear cut.

Pimm:
> Another likely example is a comparison check which will not cause a problem until this PR is merged.
That is a valid and compelling argument.

The point that I was trying to make, is that semver is supposed to guarantee that everything continues to function as expected for any non-major update. The second a payment method stops working, existing applications break, even if they don't update this client at all. The semver guarantee doesn't work in that scenario. As such, I am not sure whether it's a bad thing to break the semver rules in such a case.

Pimm (approved):

janpaepke:
I think I got your point, but as I said, if all they do is checks, the application won't break, because it simply won't receive that payment method anymore...

Pimm:
Very true, Jan.

PR #377, opened 2024-09-19
Change path of the types in package.json

Pimm:
It seems that the TypeScript compiler previously put the type definitions into `dist/types/src`, but now puts them in `dist/types`. This may have happened [during our recent upgrade of TypeScript](https://github.com/mollie/mollie-api-node/pull/360).

This PR updates package.json to use the new path.

Please do check my work!

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "main": "dist/mollie.cjs.js",
   "module": "dist/mollie.esm.js",
   "jsnext:main": "dist/mollie.esm.js",
-  "types": "dist/types/src/types.d.ts",
+  "types": "dist/types/types.d.ts",
   "engines": {
     "node": ">=8"
   },


janpaepke:
I tested your theory by building with the ts version from the previous version and comparing the dist folder structure.
It seems you are indeed correct.

The newer ts version does omit the `src` in the path, which tbh makes sense from a structure point of view.

This is a really good find! I wonder how we can better protect against something like this in the future. Maybe add [tsd](https://www.npmjs.com/package/tsd) to the testing routine?

This subltle change will also have consequences for people currently importing types from deep links, which I know you shouldn't do, [but people are definitely doing](https://github.com/mollie/mollie-api-node/issues/356).
This time I'm on the fence wether or not this should be considered breaking, but it doesn't matter much, since this part of the major.

We might still want to document it in the migration/release notes?

janpaepke (approved):

Issue #273, opened 2022-05-02
Split payments


JFGHT:
Hello there.

Does the node api support [split payments](https://docs.mollie.com/connect/splitting-payments)?

Thanks.

Pimm:
Although not officially supported, it should work:
```javascript
mollieClient.payments.create({
	amount: { currency: 'EUR', value: '10.00' },
	routing: [
		{
			amount: { currency: 'EUR', value: '2.50' },
			destination: { type: 'organization', organizationId: 'me' }
		},
		{
			amount: { currency: 'EUR', value: '3.50' },
			destination: { type: 'organization', organizationId: 'org_23456' }
		},
		{
			amount: { currency: 'EUR', value: '4.00' },
			destination: { type: 'organization', organizationId: 'org_56789' }
		}
	],
	…
});
```
(Note that ‒ as `routing` does not exist according to the type definitions ‒ the TypeScript compiler will complain if you're using it.)

Once this feature moves out of closed beta, official support to this client should be added.

Pimm:
Splitting payments has moved out of beta. This can be added to the client.

janpaepke:
This has actually already been resolved in #402, so it is part of release [v4.2.0](https://github.com/mollie/mollie-api-node/releases/tag/v4.2.0).

PR #383, opened 2024-09-27
Add support for split payments

janpaepke:
fixes https://github.com/mollie/mollie-api-node/issues/273

Currently the destination must be of type `organization`.
I still made the type an enum for consistency reasons and for potential later extensions.

I also decided to make the only currently available interface type for `destination` a type property fixed to 'organization', since the `organizationId` only makes sense in this context. Other types could then later be added as a union, creating a discriminated union.


mahdi-ko (approved):

janpaepke:
This has actually already been resolved in https://github.com/mollie/mollie-api-node/pull/402, so it is part of release [v4.2.0](https://github.com/mollie/mollie-api-node/releases/tag/v4.2.0).

PR #396, opened 2025-01-11
Added billing and shipping address and line items for Payments.

joosttk:


janpaepke:
Thank you for your PR.

We are aware of this discrepancy (there are more) and are currently working on resolving them.
The reason this isn't easy is that there are various inconsistencies in the API Docs regarding which parameters are optional and which are not.

Aside from lacking test updates your PR struggles with the same issue and just making everything optional isn't in line with this SDKs philosophy of type safety (just as an example `lines.recurring.interval` is optional in your code, but definitely isn't in the API).

We are currently working with mollie to resolve these issues ASAP.
As soon as we do we might update this PR or close and create a new one.

janpaepke:
replaced by https://github.com/mollie/mollie-api-node/pull/397

Issue #395, opened 2024-12-11
Assertion Fails when Retrieving a Single Chargeback Payment


theresa-de-ocampo:
## Reproduction Steps
1. Use `@mollie/api-client` version `4.1.0`.
2. Run the following code.
```javascript
const dotenv = require("dotenv").config();
const { createMollieClient } = require("@mollie/api-client");
const mollieClient = createMollieClient({ apiKey: process.env.MOLLIE_API_KEY });

async function run() {
  if (dotenv.error) {
    throw dotenv.error;
  }

  try {
    const chargeback = await mollieClient.paymentChargebacks.get('chb_seinif', {
      paymentId: 'tr_jv3c9soPN6'
    });
    console.dir(chargeback, { depth: null });
  } catch (error) {
    console.dir(error, { depth: null });
  }
}

run();
```

3. Check the logged errors.
```
ApiError: The refund id appears invalid: chb_seinif (unexpected format)
    at assertWellFormedId (C:\mollie-demo\node_modules\@mollie\api-client\dist\mollie.cjs.js:1933:11)       
    at PaymentChargebacksBinder.get (C:\mollie-demo\node_modules\@mollie\api-client\dist\mollie.cjs.js:3499:5)
```

## Additional Context
1. I think the `resource` value passed to `assertWellFormedId` is incorrectly set to `refund` instead of `chargeback`.

![image](https://github.com/user-attachments/assets/28c12e5c-edf2-4f19-b9e1-8f713032e6a3)

2. When `checkId` is called, it will return `false` since  `prefixes[resource]` gets evaluated to `re_`, but the value that was passed starts with `chb_`  causing `assertWellFormedId` to throw an error.

![image](https://github.com/user-attachments/assets/3a372d96-e500-452e-bf35-b1ceb3ddfc12)

janpaepke:
Thank you for the very well written issue.

I agree with your assessment and proposed fix.
Would you care to create a pr?

PR #400, opened 2025-01-27
Fix: faulty id assertion

janpaepke:
This resolves https://github.com/mollie/mollie-api-node/issues/395.

The incorrect assertion was used for chargebacks.

supersedes #399


edorivai (approved):
LGTM

Issue #390, opened 2024-11-09
Missing Typescript properties in Payment model


PaquitoSoft:
Hello 👋 

We're starting to consume Mollie APIs and I think I'm missing some properties in the Payment model typescript definition.

Based on the API documentation ([reference](https://docs.mollie.com/reference/get-payment)), a payment can have these optional properties:
- lines
- billingAddress
- shippingAddress

However the Typescript type describing this model does not contain those properties ([reference](https://github.com/mollie/mollie-api-node/blob/master/src/data/payments/data.ts#L22)).

Are we missing something?

Thanks.

janpaepke:
Especially since payments should replace the orders API, these are very sensible additions.
As these are only type additions, I am confident, we can add them fairly soon.

In the meantime, you can get around Typescript complaining by spreading them into your `payment.create` options like this:
```
lines = {
  [...]
}
client.payments.create({
	amount: {
		value: 100
		currency: "EUR",
	},
	...{ lines }, // tmp syntax, until mollie types included this property
});
```
This way you don't have to disable ts checking for the whole object.

PaquitoSoft:
Ok. Thanks for your reply 🙇 

Should we keep this opened until the types are added to the package?

fjbender:
> Should we keep this opened until the types are added to the package?

Yes! :)

carderne:
This seems to also apply to the `statusReason` field on the returned `Payment` object as well?

janpaepke:
@carderne, thanks for pointing this out. we will include it in the updates.

mahdi-ko:
`routing` is another missing field

janpaepke:
resolved in v4.2.0

PR #402, opened 2025-01-27
Feat: Update Interface Types

janpaepke:
This resolve several type discrepancies between the client and the mollie API.

Since the official docs are too inconsistent to be used as source of truth, the actual optional-ness of these properties was aligned directly with mollie.
No properties were removed and Types should be wider, not narrower, making this a backwards compatible change.

The only exception I could find was the `address.postCode`, which was defined to be required, when it is actually optional, since not all countries have a post code.
This change will concern the organization.address and order.address.
I don't believe this warrants a major version release.

This resolves:
- https://github.com/mollie/mollie-api-node/issues/390
- https://github.com/mollie/mollie-api-node/pull/396

Supersedes #397

### DIFF
--- a/src/binders/payments/parameters.ts
+++ b/src/binders/payments/parameters.ts
@@ -1,11 +1,25 @@
-import { type Address, type Amount, type PaymentMethod } from '../../data/global';
-import { type Issuer } from '../../data/Issuer';
+import { type Amount, type PaymentMethod } from '../../data/global';
 import { type PaymentData, type PaymentEmbed, type PaymentInclude } from '../../data/payments/data';
 import type MaybeArray from '../../types/MaybeArray';
 import { type IdempotencyParameter, type PaginationParameters, type ThrottlingParameter } from '../../types/parameters';
 import type PickOptional from '../../types/PickOptional';
 
-export type CreateParameters = Pick<PaymentData, 'amount' | 'description' | 'redirectUrl' | 'cancelUrl' | 'webhookUrl' | 'customerId' | 'mandateId'> &
+export type CreateParameters = Pick<
+  PaymentData,
+  | 'amount'
+  | 'description'
+  | 'redirectUrl'
+  | 'cancelUrl'
+  | 'webhookUrl'
+  | 'customerId'
+  | 'mandateId'
+  | 'lines'
+  | 'shippingAddress'
+  | 'billingAddress'
+  | 'routing'
+  | 'issuer'
+  | 'restrictPaymentMethodsToCountry'
+> &
   PickOptional<PaymentData, 'locale' | 'metadata' | 'sequenceType' | 'captureMode' | 'captureDelay'> & {
     /**
      * Normally, a payment method screen is shown. However, when using this parameter, you can choose a specific payment method and your customer will skip the selection screen and is sent directly to
@@ -45,6 +59,9 @@ export type CreateParameters = Pick<PaymentData, 'amount' | 'description' | 'red
      * @see https://docs.mollie.com/reference/v2/payments-api/create-payment?path=applePayPaymentToken#apple-pay
      */
     applePayPaymentToken?: string;
+    /**
+     * @deprecated use billingAddress.email instead
+     */
     billingEmail?: string;
     /**
      * The date the payment should expire, in `YYYY-MM-DD` format. **Note:** the minimum date is tomorrow and the maximum date is 100 days after tomorrow.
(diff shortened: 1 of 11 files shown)


src/binders/payments/parameters.ts
```diff
@@ -163,13 +171,12 @@ export type CreateParameters = Pick<PaymentData, 'amount' | 'description' | 'red
...
 
 export interface GetParameters {
-  include?: PaymentInclude;
+  include?: MaybeArray<PaymentInclude>;
   embed?: MaybeArray<PaymentEmbed>;
   testmode?: boolean;
 }
 
 export type PageParameters = PaginationParameters & {
-  profileId?: string;
```
janpaepke, line 172: see https://github.com/mollie/mollie-api-node/pull/397#discussion_r1928421828

src/binders/payments/parameters.ts
```diff
@@ -54,27 +71,12 @@ export type CreateParameters = Pick<PaymentData, 'amount' | 'description' | 'red
...
      */
     dueDate?: string;
-    /**
-     * The card holder's address details. We advise to provide these details to improve the credit card fraud protection, and thus improve conversion.
-     *
-     * If an address is provided, then the address has to be in a valid format. See the address object documentation for more information on which formats are accepted.
-     *
-     * @see https://docs.mollie.com/reference/v2/payments-api/create-payment?path=billingAddress#credit-card
-     */
-    billingAddress?: Address;
```
janpaepke, line 64: this is not actually removed, but picked from data (see above)

src/binders/payments/parameters.ts
```diff
@@ -54,27 +71,12 @@ export type CreateParameters = Pick<PaymentData, 'amount' | 'description' | 'red
...
     /**
      * The card token you got from Mollie Components. The token contains the card information (such as card holder, card number, and expiry date) needed to complete the payment.
      *
      * @see https://docs.mollie.com/reference/v2/payments-api/create-payment?path=cardToken#credit-card
      */
     cardToken?: string;
-    shippingAddress?: Address & {
```
janpaepke, line 71: this is not actually removed, but picked from data (see above)

src/data/global.ts
```diff
@@ -102,7 +120,13 @@ export interface Address {
...
    * @see https://docs.mollie.com/overview/common-data-types?path=postalCode#address-object
    */
-  postalCode: string;
+  postalCode?: string;
+  /**
+   * The customer's email address.
+   *
+   * @see https://docs.mollie.com/overview/common-data-types?path=email#address-object
+   */
+  email?: string;
```
janpaepke, line 129: @fjbender `email` is missing from the [common type definition](https://docs.mollie.com/reference/common-data-types?path=email#address-object), but I think that's a mistake, as it is included where the address is used (payments/orders)

src/data/global.ts
```diff
@@ -102,7 +120,13 @@ export interface Address {
    *
    * @see https://docs.mollie.com/overview/common-data-types?path=postalCode#address-object
    */
-  postalCode: string;
+  postalCode?: string;
```
janpaepke, line 123: this is the only **potentially** breaking change, as for this was used for orders and is actually optional.
This means users, who consumed order.billingAddress.postalCode may now get a type error.

edorivai: I agree that this particular change is more fitting for a `bugfix` version bump, than a `major`, because the original type definition (required instead of optional) was actually wrong. So if consumers of this library rely on `postalCode` not being optional, that could actually mean runtime errors on their end.

Forcing them to fix TS errors due to `postalCode` being optional is actually a good thing for all parties involves, IMO. So my vote too goes to _not_ warranting a major version bump.

src/data/orders/data.ts
```diff
@@ -221,15 +221,6 @@ export enum OrderStatus {
   pending = 'pending',
 }
 
-export interface OrderAddress extends Address {
```
edorivai, line 224: Did we drop accidentally drop
- `organizationName`
- `email`
- `phone`

by this refactor?

janpaepke: good catch. change reverted.

edorivai (approved):
LGTM from a code standpoint. @fjbender I know you synced with @janpaepke in terms of inconsistencies between actual API and API documentation. Would you mind giving this PR a once-over to confirm that these changes are in fact in line with whatever you discussed with Jan?

src/data/global.ts
fjbender: It's likely a documentation bug as the common type definition is not auto-generated.

fjbender (approved):

PR #403, opened 2025-01-27
Fixed and deprecated payment.isRefundable

janpaepke:
The `isRefundable` helper on a payment actually never returned `false`, because the strict `null` comparison is incorrect.
If `amountRemaining` is not provided (as it is for non-refundable payments), it is `undefined` rather than `null`.

This is also reflected in `canBeRefunded`, which is correct and should be used instead.
To not make this a breaking change, I fixed the behaviour and deprecated the method.

supersedes #385


edorivai (approved):
LGTM

PR #405, opened 2025-02-11
FIX: search params in returned links break iteration

### LINKED ISSUE #404: `payment.getRefunds()` doesn't seem to respect `testmode` properly
...and maybe ignores some general ApiClient properties, too?

Given you have a payment `tr_deadbeef` with a refund `re_decafbad`, and your implementation uses Organizational Access Tokens. You build the following:

```js
const { createMollieClient } = require('@mollie/api-client');

const mollieClient = createMollieClient({ apiKey: 'access_mahsecretkey' });

(async () => {
    try {
        const payment = await mollieClient.payments.get('tr_deadbeef', { testmode: true });
        const refunds = await payment.getRefunds();
        const refund = refunds.find( (refund) => refund.id === 're_decafbad' );
    }
    catch (error) {
        console.error(error);
    }
})();
```

You receive

```
ApiError: Invalid value for testmode parameter. Testmode should be string "true" or string "false".
    at ApiError.createFromResponse (/Users/florian/Dev/testbla/node_modules/@mollie/api-client/dist/mollie.cjs.js:343:12)
    at /Users/florian/Dev/testbla/node_modules/@mollie/api-client/dist/mollie.cjs.js:946:22
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/Users/florian/Dev/testbla/node_modules/@mollie/api-client/dist/mollie.cjs.js:174:24)
    at _next (/Users/florian/Dev/testbla/node_modules/@mollie/api-client/dist/mollie.cjs.js:193:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  field: 'testmode',
  statusCode: 422,
  idempotencyKey: undefined,
  title: 'Unprocessable Entity',
  links: {
    documentation: {
      href: 'https://docs.mollie.com/overview/handling-errors',
      type: 'text/html'
    }
  },
  [Symbol(Symbol.toStringTag)]: 'ApiError'
}
```

Because I am a curious person but at the same time have little to no clue about NodeJS, I tried to increase observability by pumping all traffic through a proxy, passing `apiEndpoint: 'https://mytranparentproxy.invalid/v2'` to the `createMollieClient`. The proxy passes everything through to the Mollie API. For the `payments.get` call I can observe the call being executed, but for everything after that not.

I would expect the refunds finder to a) work properly with the `testmode` get parameter and b) respect additional parameters I shoved into the `createMollieClient`.

Versions: `Node/v21.7.2 Mollie/4.1.0`

janpaepke:
Fixes #404 

`buildURL` expects a pathname (or originAndPathname) to be passed, which does not contain any search parameters.

If it does, however, the result will contain a broken url.
This happened in the example of the `getRefunds` helper, which used the returned URL which included ?testmode and appended limit params. The resulting URL looked like this:
`https://api.mollie.com/v2/payments/tr_paymentID/refunds?testmode=true?limit=128`

The solution was to parse the URL into its pathname and query components, before passing it into `.iterate`.

This issue concerned all helpers, which returned iterators.
Helpers, which only used `.get` did (accidentally) work and could have remained untouched, but I still updated them for consistency.

My initial impulse was to also update `buildUrl`, which caused the concatenation issue.
There are two options: allow existing SearchParams in the pathname and merge them or throw if there are params in the pathname.
I was leaning towards the latter, as the whole client consistently handles pathnames and query params separately.

Ultimately I decided to leave it as is, since any change (merge or throw) could lead to unexpected behaviour for users, if any networkClient call using it got overlooked.
That would make it a breaking change.

We might want to keep this in mind for the next major, though.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         node-version: [14.x, 18.x, 20.x]


edorivai (approved):
Code LGTM.

I agree with the approach of this PR; fixing all the callsites to ensure query params are broken out before calling the networkClient, and not making any potentially breaking changes yet.

In terms of future changes, we discussed
1. Throwing if the pathname contains query params
2. Parsing the pathname and merging query params if the pathname contains any

I have strong feelings against (1), because if a wrong pathname is passed, this is likely an internal bug on our part, not something the user did wrong. And in that case I wouldn't want to introduce runtime failures in downstream apps.

I can see (2) working. If we want to consider it, we should outline possible scenarios where merging could potentially break or cause unexpected behavior.

PR #406, opened 2025-02-12
NTR: Update github actions image to ubuntu-latest

fjbender:
Github Actions will deprecate the ubuntu-20.04 image, we should always run on ubuntu-latest.


janpaepke (approved):
lgtm

PR #411, opened 2025-03-17
add `organizationName` to Address Object

### LINKED ISSUE #410: organizationName missing from Address type
Hi team,

seems like we missed a property with the recent overhaul from #390 . `organizationName` is still missing from the `Address` type. 

Image

This [workaround](https://github.com/mollie/mollie-api-node/issues/390#issuecomment-2467604847) still works, in case anyone is blocked by this.

Can you take a look please?

janpaepke:
Field can be found in details for get payment: https://docs.mollie.com/reference/get-payment

The docs are still not consistent on which fields are `required` and `optional`, but here it's pretty clear: The docs say it is only included 'if the adressee is an organization'.

fixes #410

### DIFF
--- a/src/data/global.ts
+++ b/src/data/global.ts
@@ -105,6 +105,12 @@ export interface Address {
    */
   familyName?: string;
   /**
+   * The name of the organization, in case the addressee is an organization.
+   *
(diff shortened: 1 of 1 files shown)


edorivai (approved):
LGTM

Issue #387, opened 2024-09-30
Pipelines always failing fo rexternal PRs


janpaepke:
As pointed out here https://github.com/mollie/mollie-api-node/releases/tag/untagged-fa5121ce368cb7916910, all pipelines run against external PRs now fail, since they do not have access to the API test key used for integration tests.

This key was not meant to be a long term solution anyway. The idea was to run the tests against the actual API locally and record the responses as snapshots. Then in the pipelines only run tests against those snapshots.

To fix this we can either work to implement this test-flow now or create a separate pipline workflow for (external?) pull requests.
One obvious change would be to exclude the integration tests in those.

The downside here would be that failing tests would only appear once merged.

On the other hand this would also be a problem with the obove mentioned flow of generating snapshots: Any external PR could not include any integration tests for their code changes and the pipelines would also not verify wether or not the changes would break existing (integration) tests.

How do we want to approach this?

edorivai:
I propose:
1. skip integration tests that require an API key when the PR comes from a forked repo
2. ensure that forked repos merge against a different branch than `master` (e.g. `staging` or something). This way we can review and check that all non-integration tests pass. Subsequently, we can make an internal PR from `staging` to `master`, which can run integration tests using secrets from this main repo.

Alternatively: external maintainers could add _their own API Key_ to the secrets on their forked repo. This shifts the burden to the external devs, but worth considering.

Pimm:
> The idea was to run the tests against the actual API locally and record the responses as snapshots. Then in the pipelines only run tests against those snapshots.

Indeed. It also means local tests work out-of-the-box.

In my vision, "external" contributors would, as @edorivai suggests, generate snapshots using their own test API key.

> Any external PR could not include the integration tests for their code changes and the pipelines would also not definitively verify wether or not the changes would break existing (integration) tests. For example they might work against the snapshots but fail agains the actual API results, which might be different with the changed code.

I don't think I get your concern. If an "external" contributor creates a PR which changes the requests going to the Mollie API, the test will fail because the snapshots don't match (and they might have to generate new snapshots). If the existing snapshots still match, that should give us confidence that the changes don't break anything, and should still work as expected when touching the real Mollie API.

The snapshots ‒ by their nature ‒ reflect the real Mollie API, so they should give us the same confidence as a test against the real Mollie API.

Could you elaborate?

PR #329, opened 2023-09-08
Adds the 'new' terminal API

Jeffrey-Zutt:
This PR adds the 'new' terminal API.
As discussed in issue: https://github.com/mollie/mollie-api-node/issues/328

Please let me know if you'd like anything changed :-)

Issue #328, opened 2023-09-07
The new 'Terminals' API


Jeffrey-Zutt:
Hello!

First off all, thanks for maintaining this great library :-).

I was wondering if there are any plans of integrating the new 'terminals' API into this library.
https://docs.mollie.com/reference/v2/terminals-api/overview

Happy to help and raise a PR on it.

fjbender:
Thanks for reaching out! We're always happy to take contributions :)

Once the `resource` issue on the Terminal API is covered, let's take another look at this.

Jeffrey-Zutt:
Yeah I noticed that too, the `resource` property is missing from the actual API response right? 
(Even though it's documented here: https://docs.mollie.com/reference/v2/terminals-api/get-terminal#response)

Do you happen to have any idea when this issue will be covered? (Ballpark?)

fjbender:
I have just been informed that one of our fantastic engineers has fixed it already :) Should be available very shortly.

Jeffrey-Zutt:
Awesome! I'll keep you posted :-)

Jonas4Mollie:
The resource parameter is now part of the API response, so you should be able to move ahead!

iFlyinq:
Any updates on this one?

janpaepke:
addition is planned for 4.2.0

PR #413, opened 2025-03-17
Adds the 'new' terminal API

janpaepke:
This PR adds the 'new' terminal API.
As discussed in issue: #328

This replaces PR #329, because the pipelines fail, if a PR is not issued from this repo.
See #387 for details.


src/binders/payments/parameters.ts
```diff
@@ -77,6 +77,13 @@ export type CreateParameters = Pick<
      * @see https://docs.mollie.com/reference/v2/payments-api/create-payment?path=cardToken#credit-card
      */
     cardToken?: string;
+    /**
+     * The unique identifier used for referring to a terminal. This ID is used for assigning the payment to a specific terminal and it can be retrieved via List terminals. For more information about
+     * point-of-sale payments, please check our guide point-of-sale payments.
+     *
+     * @see https://docs.mollie.com/reference/v2/payments-api/create-payment?path=terminalId#point-of-sale
+     */
+    terminalId?: string
```
janpaepke, line 86: I cannot find this parameter in [create-payment](https://docs.mollie.com/reference/create-payment).

Is this a doc issue or does it not exist (anymore)?

src/data/payments/data.ts
```diff
@@ -828,6 +829,50 @@ export interface PaysafecardDetails {
   customerReference: string;
 }
 
+export interface PointOfSaleDetails {
```
janpaepke, line 832: In the [method specific details for POS](https://docs.mollie.com/reference/extra-payment-parameters#point-of-sale-1) I can only see `terminalId`.

Are the other properties also included? Are they all non-optional?

janpaepke (commented):
This PR is functionally ready.
@fjbender, please confirm type specific questions (see inline comments).

Also the integration tests require at least one terminal to exist in the test account.
Since they cannot be created through the API, you might have to manually set one up in the test account.

src/binders/payments/parameters.ts
fjbender: It's defined here: https://docs.mollie.com/reference/extra-payment-parameters#point-of-sale, we moved all the method-specific parameters to that section. Typing it as string is fine, we might want to update the `@see` link.

janpaepke: this might be a misunderstanding.
Within the package 'parameter' refers to request parameters, that can be sent to the API.
'data' refers to response data, that is returned by the API.

I believe the link above refers to response data. specifically `payment.details`.
In fact there is a separate comment below, regarding this.

Here I'm specifically asking, if `terminalId` is allowed as a request parameter for `payment.create`, since it does not appear in the [docs](https://docs.mollie.com/reference/create-payment).

src/data/payments/data.ts
janpaepke: To clarify: This relates to line 839 to 875.
These were part of the original PR for this update, but I cannot confirm where these properties came from or if they are accurate.

fjbender: - Details
	- terminalId
	- cardNumber
	- cardAudience
	- cardLabel
		- Possible values limited to Visa or Mastercard for terminal transactions. Maestro is currently labeled as Mastercard
	- cardCountryCode
	- cardFingerprint

src/binders/payments/parameters.ts
janpaepke: @fjbender still need your clarification on this one, pls. :)

tests/integration/terminals.test.ts
```diff
@@ -0,0 +1,20 @@
+import dotenv from 'dotenv';
+import createMollieClient from '../..';
+
+/**
+ * Load the API_KEY environment variable
+ */
+dotenv.config();
+
+const mollieClient = createMollieClient({ apiKey: process.env.API_KEY });
```
edorivai, line 9: Just FYI:

Types of property 'apiKey' are incompatible.
        Type 'string | undefined' is not assignable to type 'string'.
          Type 'undefined' is not assignable to type 'string'.

src/binders/payments/parameters.ts
janpaepke: I just realised that you actually linked to additional create params. so terminalId IS in there.

Thanks for the clarification.

edorivai (approved):
Code LGTM

@fjbender - I think you are aware that we weren't able to test due to not having any Terminals connected to the test account, right?

PR #414, opened 2025-03-18
NTR: Add release payment authorization endpoint

janpaepke:
This adds the [release payment authorization endpoint](https://docs.mollie.com/reference/release-authorization).

Note that this endpoint currently erroneously returns an empty response (rather than an empty json object).
Currently the package only expected empty responses for 204 status codes.
This is in line with the intentions of REST, but since this is convention rather than a strict requirement I adapted our NetworkClient to not fail in these scenarios.

Regardless I will recommend to mollie to update their endpoint for consistency.

### DIFF
--- a/src/communication/NetworkClient.ts
+++ b/src/communication/NetworkClient.ts
@@ -81,14 +81,24 @@ const throwApiError = run(
 async function processFetchResponse(response: ResponseWithIdempotencyKey) {
   // Request was successful, but no content was returned.
   if (response.status == 204) return true;
-  // Request was successful and content was returned.
-  const body = await response.json();
-  if (Math.floor(response.status / 100) == 2) {
-    return body;
+  const body = await response.text();
+  const isSuccessStatus = Math.floor(response.status / 100) == 2;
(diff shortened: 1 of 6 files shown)


src/communication/NetworkClient.ts
```diff
@@ -81,14 +81,24 @@ const throwApiError = run(
...
-  // Request was successful and content was returned.
-  const body = await response.json();
-  if (Math.floor(response.status / 100) == 2) {
-    return body;
+  const body = await response.text();
+  const isSuccessStatus = Math.floor(response.status / 100) == 2;
+  // this should have been a 204, but we'll be lenient...
+  if (body.length === 0 && isSuccessStatus) return true;
+  // Parse the response body as JSON.
+  // We could check the content-type header first here, but we always expect thie mollie API to return JSON.
```
edorivai, line 89: typo
```suggestion
  // We could check the content-type header first here, but we always expect the mollie API to return JSON.
```

edorivai (approved):
LGTM - minor typo in comment, nothing blocking

PR #416, opened 2025-03-20
Add payments pagination support for payment links

TimmyGuy:
Introduced a new `payments` method in `PaymentLinksBinder` to retrieve paginated payments for a given payment link. Updated type definitions and test cases to ensure functionality and coverage, including fixtures for API responses.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,5 @@
   },
   "files": [
     "dist/**/*"
-  ],
-  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -78,5 +78,6 @@
   },
   "files": [
     "dist/**/*"
-  ]
+  ],
+  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
 }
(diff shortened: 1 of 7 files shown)


TimmyGuy:
You are right @janpaepke, I've altered the code to use the helper. Looks much cleaner and more how the rest of the API works. The extra request is a small price. For now it uses the self link with /payments appended, mainly so I can use it in my project. If the API supports the payments link, it can be easily changed!

TimmyGuy:
I've also added the missing methods for the payment-links api. Should now cover the entire paymentlinks endpoint.

janpaepke:
I think this is a really elegant solution and as you say it can easily be adapted later on.

However, please do not mix PRs. This should be only about adding the payment retrieval.
As a matter of fact the Payment Link types are already taken care of here #419.

TimmyGuy:
Sure, ill make a separate PR for it.

janpaepke:
no need for a separate PR. as I said, the types are already done in #419 

Is this one done from your side?

TimmyGuy:
Haha, oops, already reverted the commit. Luckily git has version history. Yeah this one is all done from my side.

janpaepke:
oooh. Actually I overlooked the `update` and `delete` methods. Yeah an additional PR would actually make sense for these.

thanks :)

janpaepke:
this PR is replaced by #421 because of #387

PR #419, opened 2025-03-20
Added missing payment link params

### LINKED ISSUE #415: `minimumAmount` on Payment Links API
We have added a new field for payment links creation: `minimumAmount`. It's an amount object, so it consists of a currency and the value represented as string. Docs [here](https://arc.net/l/quote/dzjejbfj).

janpaepke:
There were several missing types in payment links api.

This fixes #415


edorivai (approved):
LGTM, cross check documentation with these changes

PR #422, opened 2025-03-20
Add `update` and `delete` methods for payment links 

janpaepke:
Introduce `update` and `delete` functionality to the `PaymentLinksBinder`. The `update` method allows modifying payment link details, while the `delete` method enables removal of payment links. Also, extend `parameters.ts` to include `UpdateParameters` for supporting these changes.

replaces #420


edorivai (approved):
LGTM

PR #424, opened 2025-03-24
Remove `updatedAt` field from payment links

janpaepke:
As discussed [here](https://github.com/mollie/mollie-api-node/issues/415#issuecomment-2747217823), the `updatedAt`  field in payment links always returns null should not exist.

Since it has been in the package for more than 3 years I decided on deprecating it, rather than removing it straight away.


edorivai (commented):
lgtm

edorivai (approved):

PR #429, opened 2025-04-15
Add missing method specific parameters

janpaepke:
This adds several [method specifc params](https://docs.mollie.com/reference/extra-payment-parameters#paypal), which are currently missing.

It also updates the reference links and doc texts to the current version(s).

In the long run we might want to improve this using discriminated unions, but for now this consistent with past implementations.


edorivai (approved):
LGTM

PR #432, opened 2025-04-22
add paybybank payment method

hreinberger:
This PR adds the new paybybank payment method

Info see here: https://www.mollie.com/payments/pay-by-bank


janpaepke:
replaced and closed because of because of #387.

PR #433, opened 2025-04-24
add paybybank payment method

janpaepke:
This PR adds the new paybybank payment method

Info see here: https://www.mollie.com/payments/pay-by-bank

This PR replaces #432 because of #387.


edorivai (approved):
LGTM

PR #439, opened 2025-06-06
Update parameters.ts

n2k3:
Added missing `cancelUrl` from `PaymentData`


janpaepke:
Please always create an issue first, so we can discuss together wether or not this is the best solution.
This property is definitely not the only one missing here.

This is the corresponding issue for tracking: https://github.com/mollie/mollie-api-node/issues/441

I will close this issue, since we will have to recreate it anyway, due to #387

PR #461, opened 2025-11-26
feat: add OAuth API support

### LINKED ISSUE #456: Add OAuth API support
## Description

Add support for the Mollie OAuth API endpoints to enable OAuth token management for Mollie Connect applications.

## API Reference
https://docs.mollie.com/reference/oauth-api

### 1. Generate tokens
- **Method**: POST
- **Endpoint**: `/oauth2/tokens`
- **Purpose**: Exchange authorization code for access token, or refresh an access token
- **Request body**:
  - `grant_type`: 'authorization_code' | 'refresh_token' (required)
  - `code`: string | null (for authorization_code grant type)
  - `refresh_token`: string | null (for refresh_token grant type)
  - `redirect_uri`: string | null
- **Response**: Token object with:
  - `access_token`: string
  - `refresh_token`: string
  - `expires_in`: integer
  - `token_type`: 'bearer'
  - `scope`: string

### 2. Revoke tokens
- **Method**: DELETE
- **Endpoint**: `/oauth2/tokens`
- **Purpose**: Revoke an access token or refresh token
- **Request body**:
  - `token_type_hint`: 'access_token' | 'refresh_token' (required)
  - `token`: string (required)
- **Response**: Empty (204)

## Implementation Notes

- Both endpoints use OAuth client credentials (Basic Auth) instead of API keys
- The Authorize endpoint (browser redirect) does not need SDK implementation
- These endpoints are critical for Mollie Connect OAuth flow
- Revoking a refresh token revokes all associated access tokens

janpaepke:
## Summary

Adds support for the OAuth API endpoints, enabling applications to generate and revoke OAuth tokens programmatically.

## Changes

- **Added OAuth data model** in `src/data/oauth/` with Token response type and enums (GrantType, TokenType)
- **Implemented `OAuthBinder`** with `create()` for token generation and `revoke()` for token revocation
- **Added `rawRequest` to NetworkClient** as an escape hatch for endpoints that don't follow standard v2 conventions (OAuth uses Basic Auth, form-urlencoded body, and `/oauth2/tokens` endpoint)
- **Exported OAuth types** in `types.ts` and enums in `createMollieClient.ts`
- **Added unit tests** verifying request format (Basic Auth header, form-urlencoded body, discriminated union grant types)

The implementation uses discriminated unions on `grant_type` for type-safe token generation (authorization_code vs refresh_token grants).

## API Details

The OAuth API provides:
- Token generation via authorization code exchange
- Token generation via refresh token
- Token revocation for both access and refresh tokens

## Note

Unable to add integration tests as this endpoint requires OAuth application credentials. Unit tests verify the request format but cannot verify actual API behavior.

**This will be released as a beta first.**

## References

- Closes #456
- [OAuth API docs](https://docs.mollie.com/reference/oauth-api)
- [Generate tokens](https://docs.mollie.com/reference/oauth-generate-tokens)
- [Revoke token](https://docs.mollie.com/reference/oauth-revoke-token)

### DIFF
--- a/tests/unit/errors.test.ts
+++ b/tests/unit/errors.test.ts
@@ -1,4 +1,4 @@
-import { MollieApiError, MollieClient, PaymentCreateParams } from '../..';
+import { MollieApiError, MollieClient, OAuthGrantType, PaymentCreateParams } from '../..';
 import NetworkMocker, { getApiKeyClientProvider } from '../NetworkMocker';
 
 describe('errorHandling', () => {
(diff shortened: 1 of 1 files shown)


src/binders/oauth/OAuthBinder.ts
```diff
@@ -0,0 +1,58 @@
+import type NetworkClient from '../../communication/NetworkClient';
+import type Token from '../../data/oauth/data';
+import renege from '../../plumbing/renege';
+import type Callback from '../../types/Callback';
+import { type CreateParameters, type RevokeParameters } from './parameters';
+
+const endpoint = 'https://api.mollie.com/oauth2/tokens';
+
+const generateToken = (clientId: string, clientSecret: string) => Buffer.from(`${clientId}:${clientSecret}`).toString('base64');
```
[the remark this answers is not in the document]
janpaepke: Fixed in ced0899. Renamed to `encodeBasicAuthCredentials`.

src/binders/oauth/OAuthBinder.ts
```diff
@@ -0,0 +1,58 @@
+import type NetworkClient from '../../communication/NetworkClient';
+import type Token from '../../data/oauth/data';
+import renege from '../../plumbing/renege';
+import type Callback from '../../types/Callback';
+import { type CreateParameters, type RevokeParameters } from './parameters';
+
+const endpoint = 'https://api.mollie.com/oauth2/tokens';
```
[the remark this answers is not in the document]
janpaepke: Fixed in ced0899. Made endpoint a constructor parameter with the default value.

src/binders/oauth/parameters.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Discriminated union based on `grant_type`:
+ * - `authorization_code`: requires `code`
+ * - `refresh_token`: requires `refresh_token`
+ */
+type GrantParameters = AuthorizationCodeGrant | RefreshTokenGrant;
+
+export type CreateParameters = OAuthCredentials &
+  GrantParameters & {
+    /**
+     * The URL the merchant is sent back to once the request has been authorized.
```
edorivai: Nit, seems that there is some additional info in the mollie docs (maybe added after you opened this PR?):

> The URL the merchant is sent back to once the request has been authorized. It must match the URL you set when registering your app.
> 
> For consecutive refresh token requests, this parameter is required only if the initial authorization code grant request also contained a `redirect_uri`.

src/communication/NetworkClient.ts
```diff
@@ -287,4 +287,15 @@ export default class NetworkClient {
     };
     return this.request(buildUrl(pathname), config);
   }
+
+  /**
+   * Low-level request method for endpoints that don't follow the standard Mollie API conventions.
+   * Allows full control over URL, headers, and body encoding while still using retry logic and error handling.
+   *
+   * @param url - Full URL (not relative to apiEndpoint)
```
edorivai: Tested in chrome that passing the full URL does what we expect - the `new URL(...)` constructor ignores the second param (which this NetworkClient) sets to the api endpoint. ✅ 

Guess the only question that leaves is whether all NodeJS versions that we want to support follow the exact same behavior.

tests/unit/resources/oauth.test.ts
```diff
@@ -0,0 +1,211 @@
...
+        const token = await bluster(mollieClient.oauth.create.bind(mollieClient.oauth))({
+          clientId: 'app_xxxxxx',
+          clientSecret: 'secret_xxxxxx',
+          grant_type: OAuthGrantType.authorization_code,
+          code: 'auth_code_xxxxxx',
+          redirect_uri: 'https://example.com/callback',
+        });
+
+        // Verify response
+        expect(token.access_token).toBe('access_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx');
```
edorivai: nit: extract these expected strings into consts, so we can be sure they're identical and we don't have to count `x`s

PR #473, opened 2026-03-16
chore: update dev dependencies

janpaepke:
## Summary

- Bump `@babel/core`, `@babel/plugin-transform-class-properties`, `@babel/preset-env`, `dotenv`, and `prettier` to latest minor versions
- Upgrade `@rollup/plugin-babel` from v6 to v7 (only drops Rollup 1 support, no other breaking changes)
- Upgrade `typescript` from v5 to v6
- Fix TS 6 breaking changes in tsconfig: explicit `moduleResolution: "node10"` with `ignoreDeprecations` flag, and explicit `rootDir` for declaration builds

## Test plan

- [x] `yarn build` passes
- [x] `yarn test` passes (53 suites, 160 tests)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,12 +47,12 @@
     "ruply": "^1.0.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.0.0-0",
-    "@babel/plugin-transform-class-properties": "^7.27.1",
-    "@babel/preset-env": "^7.28.5",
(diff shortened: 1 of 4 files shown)


edorivai (approved):
LGTM.

Note to ourselves; maybe it's time to migrate back to npm from yarn. Also I think it's better to not use lockfiles for libraries like this.

PR #474, opened 2026-03-31
feat: add sort parameter to list endpoints

janpaepke:
## Summary
- Adds `SortParameter` type (`sort?: 'asc' | 'desc'`) to `src/types/parameters.ts`
- Mixes it into `PageParameters` for all 12 list endpoints that support it per the Mollie API docs: payments, orders, customers, chargebacks, subscriptions, terminals, refunds (all), customer payments, mandates, customer subscriptions, subscription payments, and balance transfers
- Adds JSDoc to `PaginationParameters` fields

## Test plan
- [x] TypeScript build passes
- [x] Verify `sort` param is correctly passed through to the API on list calls


edorivai (approved):
LGTM

PR #475, opened 2026-04-01
feat: add TestModeParameter type to replace inline testmode definitions

janpaepke:
## Summary
- Adds `TestModeParameter` interface to `src/types/parameters.ts`, mirroring the existing `SortParameter` approach
- Replaces all inline `testmode?: boolean` definitions across 19 binder parameter files with extends/intersections of the shared type
- Empty interfaces that only contained `testmode` were converted to intersected type aliases


edorivai (approved):
LGTM, nice cleanup

PR #480, opened 2026-06-24
feat: add storeCredentials parameter to payment create and order create

fjbender:
fixes #479

### DIFF
--- a/src/binders/payments/parameters.ts
+++ b/src/binders/payments/parameters.ts
@@ -82,6 +82,15 @@ export type CreateParameters = Pick<
      */
     cardToken?: string;
     /**
+     * Whether to store the customer's payment credentials for future payments. When set to `true`, the customer's card details can be used for subsequent payments without the customer having to re-enter
+     * their card information.
+     *
+     * Requires a `customerId` to be provided.
+     *
+     * @see https://docs.mollie.com/reference/create-payment
+     */
+    storeCredentials?: boolean;
+    /**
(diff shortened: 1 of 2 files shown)


src/binders/payments/parameters.ts
```diff
@@ -81,6 +81,15 @@ export type CreateParameters = Pick<
...
     cardToken?: string;
+    /**
+     * Whether to store the customer's payment credentials for future payments. When set to `true`, the customer's card details can be used for subsequent payments without the customer having to re-enter
+     * their card information.
+     *
+     * Requires a `customerId` to be provided.
+     *
+     * @see https://docs.mollie.com/reference/create-payment
+     */
+    storeCredentials?: boolean;
```
janpaepke, line 91: The JSDoc understates the API requirements. Mollie's create-payment reference requires `customerId`, `cardToken`, **and** the `creditcard` method to be specified — not just `customerId`. A consumer trusting this doc could send `storeCredentials: true` with only a `customerId` and hit a 422.

Suggest syncing to the official wording:

```suggestion
    /**
     * Whether the card details should be stored for the customer after a successful payment. This will create a mandate for the customer, allowing for future customer present saved-card CIT payments.
     *
     * Requires `customerId`, `cardToken`, and the `creditcard` method to be specified.
     *
     * @see https://docs.mollie.com/reference/create-payment
     */
    storeCredentials?: boolean;
```

janpaepke (approved):

PR #482, opened 2026-06-25
chore(deps): update dev dependencies and patch form-data advisory

janpaepke:
## What

- Minor/patch bumps across the build & tooling chain: `@rollup/plugin-babel`, `commitizen`, `dotenv`, `prettier`, `typescript`.
- Re-resolves the transitive `form-data` to **4.0.6**, clearing a **High** CRLF-injection advisory (npm 1120743) that showed up in the *production* dependency tree via `@types/node-fetch`.
- **CI (Node matrix):** scope `--ignore-engines` to the Node 14 job — `commitizen` 4.3.2 raised its `engines.node` floor to `>=18` and was aborting the Node 14 install — and extend the test matrix to **Node 22 & 24** with `fail-fast: false` so one version's failure no longer masks the others.
- **CI (action versions):** bump `actions/checkout` and `actions/setup-node` to **v6**, and move `github/codeql-action` off the deprecated **v2** to **v4** (no v6 exists). Enable yarn dependency caching via `cache: yarn` — `setup-node` v6 narrowed automatic caching to npm only, so without this the matrix re-fetches every dependency from scratch on each job.

## Notes

- All bumps verified locally: `yarn build` ✓, unit suite (110 tests) ✓, `yarn lint` ✓, and a read-only integration smoke test ✓.
- `form-data` needed no `resolutions` entry — the existing `^4.0.4` range already permits 4.0.6, so the lockfile was simply re-resolved.
- `yarn audit --groups dependencies` (production only) now reports **0 vulnerabilities** (was 1 High).
- No major upgrades included — those are blocked by the Node 14 CI floor (and `node-fetch` 3 by its ESM-only switch); tracked separately.
- The `--ignore-engines` scoping is deliberately limited to the Node 14 job: Node 18/20/22/24 keep strict engine checks, and the workaround disappears on its own once Node 14 is dropped from the matrix.
- Action majors pinned to the latest available: `checkout` and `setup-node` v6 (their newest is v7/v6 respectively — v6 chosen as the consistent floor), `codeql-action` v4 (latest; v2 is deprecated). CI is green across the full matrix.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -18,11 +18,11 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v2
+        uses: github/codeql-action/init@v4
         with:
           config-file: ./.github/codeql-configuration.yml
           languages: javascript
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v2
+        uses: github/codeql-action/analyze@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,8 +10,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        node-version: [14.x, 18.x, 20.x]
+        node-version: [14.x, 18.x, 20.x, 22.x, 24.x]
     steps:
       - name: Checkout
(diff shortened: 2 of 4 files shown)


edorivai (approved):
Reviewed: the `--ignore-engines` scoping is done right — it's a per-job matrix conditional (`matrix.node-version == '14.x'`), not applied globally. form-data is transitive-only (re-resolved to 4.0.6, clearing the advisory); all dep bumps are dev-dep patch/minor. Confirmed the action version bumps exist: `checkout@v6`, `setup-node@v6`, `codeql-action@v4` are all real, current majors. LGTM.

PR #483, opened 2026-06-25
fix(captures): embed the payment via the `embed` query parameter

janpaepke:
## What

Captures could never embed the related payment. The Captures API embeds it through the `embed` query parameter, but the SDK sent `include`, which Mollie silently ignores — so `_embedded.payment` was never populated and `capture.getPayment()` always made a second request. The embed option was effectively dead.

## Fix (non-breaking)

- Add `embed` / `CaptureEmbed`.
- Keep `include` / `CaptureInclude` as `@deprecated` aliases that now resolve onto `embed`, so existing code keeps compiling **and** starts embedding correctly.
- Applies to payment captures (`get` / `list` / `iterate`) and settlement captures (`list` / `iterate`).

The deprecated→canonical mapping goes through a new generic `foldParameters` plumbing helper: target and source keys are both constrained to `keyof T` (a mistyped name is a compile error, not a silent no-op), the return type is a precise `Omit<T, source>`, and the key is left off the wire entirely when absent.

## Docs synced to spec

- Field and endpoint (binder-method) JSDoc refreshed against the current OpenAPI spec; dropped stale "supported methods" / legacy Orders-shipment-flow claims.
- `@see` URLs moved to the `/reference/...` format.

## Tests / verification

- Regression tests assert `?embed=payment` is actually sent — for both `embed` and the deprecated `include`, on `get` and `list` (including that the `all` / `list` aliases share the patched method).
- Unit tests for `foldParameters` (incl. the `{ to, from: undefined } -> { to }` edge).
- Full unit suite green; bundle loads and tests pass on Node 14.

## Notes / follow-ups

- `embed=payment` is taken from the spec; I couldn't reach the live API — worth a quick confirmation before release that `?embed=payment` populates `_embedded.payment`.
- `settlementAmount`: the spec omits it on the standalone capture response (it appears only on settlement-embedded captures). Left as-is (required) — flagged, not changed, since removing/narrowing it would be breaking and it is likely present in real responses.

### DIFF
--- a/src/binders/payments/captures/PaymentCapturesBinder.ts
+++ b/src/binders/payments/captures/PaymentCapturesBinder.ts
@@ -4,6 +4,7 @@ import type Capture from '../../../data/payments/captures/Capture';
 import { type CaptureData } from '../../../data/payments/captures/data';
 import alias from '../../../plumbing/alias';
 import assertWellFormedId from '../../../plumbing/assertWellFormedId';
+import foldParameters from '../../../plumbing/foldParameters';
 import renege from '../../../plumbing/renege';
 import type Callback from '../../../types/Callback';
 import Binder from '../../Binder';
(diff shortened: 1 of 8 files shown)


edorivai (approved):
Reviewed closely since this is the behavior-changing one. The `foldParameters` helper is correct: folds `include`→`embed`, handles the both-passed collision (target wins via `??`), drops absent keys so no `embed=undefined` hits the wire, and the generic typing is sound. Wired into all the claimed binder methods (payment + settlement captures inherit it). Tests assert the `?embed=payment` wire format for both the `embed` and deprecated `include` paths. Real fix — embedding was silently broken before. LGTM.

Minor (non-blocking): settlement captures have no embed-specific test, though they share the helper so a regression is unlikely.

PR #488, opened 2026-06-26
feat(methods): sync types and docs with Mollie API spec

janpaepke:
## Summary

Syncs the **Methods API** types and JSDoc against Mollie's current master OpenAPI spec. The types had drifted: a recently-added payment method and locales, a missing query parameter, and several stale or incorrect doc comments and legacy `@see` links.

## Changes

**Types (additive, non-breaking)**
- `PaymentMethod` enum += `googlepay`
- `Locale` enum += `en_GB`, `fr_LU`, `de_LU` — `en_GB` was already advertised in JSDoc but never defined in the enum, so it was unusable
- `GetParameters` += `sequenceType` query parameter (accepted by get-method per the spec)

**Documentation**
- Restored the `maximumAmount` null caveat (`null` when there is no method-specific maximum)
- Fixed the `issuers` `@see` anchor (it pointed at `path=pricing`)
- Flagged the Orders-API deprecation on the `resource` list parameter
- Added `googlepay` to `includeWallets` possible values
- Documented `list()` behaviour: online-only methods, Euro-default unless `amount` is passed
- Migrated all `@see` links from `/reference/v2/methods-api/...` to `/reference/...`

No response or required-field shapes were changed.

## Scope

Covers the Methods API: get-method, list-methods, list-all-methods. The profile-level enable/disable method operations live under the Profiles API and are out of scope for this PR.

## Docs references

- https://docs.mollie.com/reference/get-method
- https://docs.mollie.com/reference/list-methods
- https://docs.mollie.com/reference/list-all-methods

## Optional follow-up (not in this PR)

`GET /v2/methods/all` (list-all-methods — every method Mollie offers regardless of eligibility) is not currently exposed by the SDK. Adding it would be a new feature rather than a fix.


edorivai (approved):
Reviewed: additive and non-breaking. `googlepay` added to the shared `PaymentMethod` enum, new locales (`en_GB`/`fr_LU`/`de_LU`) added, and `sequenceType?` added to `GetParameters` reusing the existing `SequenceType` enum — all confirmed absent from current master, so genuinely new. Re-verified against up-to-date master: the params already `extends TestModeParameter`, so no divergence and it's MERGEABLE. LGTM.

PR #492, opened 2026-06-29
feat(balance-transfers): support metadata on connect balance transfers

janpaepke:
## Summary

Syncs the `balance-transfers` types with Mollie's OpenAPI spec for `/v2/connect/balance-transfers`.

- **`metadata`** — add the free-form `metadata` object to `BalanceTransferData` (response) and expose it as an optional field on `CreateParameters`. It is part of the connect-balance-transfer entity but was missing from the SDK, so consumers could neither send nor read it.
- **JSDoc** — restore the `amount` example from the spec and document the source-organization authentication / `balance-transfers.write` scope requirement on `create()`.

## Deliberately not changed

`statusReason` stays optional (`statusReason?`) on the response. The spec marks it `required`, but its own description ("if applicable") and observed behaviour indicate it can be absent — forcing it required would be a semantically-wrong, breaking change.

## Reference

- [Create a Connect balance transfer](https://docs.mollie.com/reference/create-connect-balance-transfer)
- [Get a Connect balance transfer](https://docs.mollie.com/reference/get-connect-balance-transfer)
- [List Connect balance transfers](https://docs.mollie.com/reference/list-connect-balance-transfers)

## Verification

- `yarn build:declarations` — passes (full type graph)
- `eslint` on touched dirs — clean
- `prettier` on edited files — clean

### DIFF
--- a/src/binders/balance-transfers/BalanceTransfersBinder.ts
+++ b/src/binders/balance-transfers/BalanceTransfersBinder.ts
@@ -21,6 +21,8 @@ export default class BalanceTransfersBinder extends Binder<BalanceTransferData, 
    * Create a balance transfer from your organization's balance to a connected organization's balance, or vice versa.
    * You can also create a balance transfer between two connected organizations.
    *
+   * To create a balance transfer, you must be authenticated as the source organization, and the destination organization must be a connected organization that has authorized the `balance-transfers.write` scope for your organization.
+   *
    * @since 4.4.0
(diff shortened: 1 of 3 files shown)


edorivai (approved):
Reviewed: `metadata` added as optional on create via `PickOptional` and as `unknown` on the response data type — consistent with how every other resource models metadata. Non-breaking. LGTM.

PR #493, opened 2026-06-29
test: group scenario suites under tests/scenarios

janpaepke:
## What

Moves three test suites out of their lone single-file top-level directories into a shared `tests/scenarios/` folder:

- `tests/paymentLinks/paymentLinks.test.ts` → `tests/scenarios/paymentLinks.test.ts`
- `tests/profiles/profiles.test.ts` → `tests/scenarios/profiles.test.ts`
- `tests/settlements/settlements.test.ts` → `tests/scenarios/settlements.test.ts`

## Why

These three were the only single-file top-level directories under `tests/` — an artifact of the original binder commit that was never migrated when the rest of the layout settled.

All three are **multi-step lifecycle tests**: access-token, helper-driven walks across a resource and its sub-resources (e.g. `settlement.getPayments()/getCaptures()/getRefunds()/getChargebacks()`). That makes them a distinct category from:

- `tests/unit/resources/` — method-focused, inline-mock unit tests
- `tests/integration/` — live-API smoke tests
- `tests/iteration/` — pagination record/replay tests

Grouping them under `tests/scenarios/` reflects what they are. The existing `tests/unit/resources/profiles.test.ts` (a genuine unit test) is untouched — no collision.

## Notes

Pure relocation, no behavioural change:

- Identical directory depth, so the relative imports (`../..`, `../getHead`, `../NetworkMocker`) are unchanged.
- nock fixtures resolve via `NetworkMocker.ts`'s location, not the test file path, so the record/replay suites (`paymentLinks`, `profiles`) still find `tests/__nock-fixtures__/*.json`.
- Git records all three as 100%-similarity renames.

All 9 tests pass.


edorivai (approved):
Reviewed: pure file move (3 renames, 100% similarity, +0/-0). Verified the jest `testRegex` still matches the new `tests/scenarios/` paths so no suites are silently dropped, and the relative imports resolve unchanged (same directory depth). LGTM.

PR #495, opened 2026-06-29
feat: add legalEntity, registrationOffice, incorporationDate to client link params

janpaepke:
Adds three organization onboarding fields to `CreateParameters` for the **Create client link** endpoint that were present in the API spec but missing from the SDK:

- `legalEntity?: string`
- `registrationOffice?: string`
- `incorporationDate?: string`

All three are request-only (`writeOnly` — not returned in the response), so they're declared inline in `parameters.ts` rather than picked from a `Data` interface. All optional, with verbatim spec descriptions and `@see` links.

### Docs reference
- [Create client link](https://docs.mollie.com/reference/create-client-link)

Found via `/sync-api client-links`.

### DIFF
--- a/src/binders/client-links/parameters.ts
+++ b/src/binders/client-links/parameters.ts
@@ -55,4 +55,24 @@ export interface CreateParameters extends IdempotencyParameter {
    * @see https://docs.mollie.com/reference/create-client-link?path=vatNumber#body-params
    */
   vatNumber?: string;
+  /**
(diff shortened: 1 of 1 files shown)


edorivai (approved):
Reviewed: all three new params (`legalEntity`, `registrationOffice`, `incorporationDate`) are optional/additive, typed as `string` (correct — no shared type exists for these codes; `incorporationDate` is `YYYY-MM-DD`), added to the right `CreateParameters`. LGTM.

PR #504, opened 2026-06-29
feat(subscriptions): reconcile types with the Mollie API spec

janpaepke:
Reconciles the **subscriptions** resource against the master OpenAPI spec (types, JSDoc, enums, `@see` links, helper wiring). Non-breaking; additive only.

## Changes

**Structure**
- Add the `_links.mandate` entry to `SubscriptionData` and a `SubscriptionHelper.getMandate()` helper, so consumers can resolve the subscription's mandate without reaching into the underscore-prefixed link.
- Add the `profileId` query parameter to the organization-wide list (`GET /v2/subscriptions`) and the subscription-payments list (`GET /v2/customers/{id}/subscriptions/{id}/payments`) — supported by the API for OAuth/organization-level credentials but previously missing from the SDK.

**Documentation**
- Restore spec constraints that had been dropped from the JSDoc:
  - `times` — test-mode subscriptions are auto-canceled after 10 payments.
  - `description` — must be unique per customer with multiple active subscriptions.
  - `webhookUrl` — receives **all** events for the subscription's payments, including failures.
  - `metadata` — automatically forwarded to the payments generated for the subscription.
- Modernize 25 `@see` references from the legacy `/reference/v2/subscriptions-api/<op>` form to the current `/reference/<op>` URLs.
- Add missing JSDoc on the new `profileId` params and the `SubscriptionsBinder.iterate` `@see`; fix a "subscriptions"→"subscription" typo in the payments binder.

## Deferred (breaking — tracked in #489)

Two response/request-shape discrepancies with the spec are intentionally left unchanged because correcting them would break consumers; documented on #489:
- `profileId` is declared as a **response** field on `SubscriptionData`, but the spec marks it `writeOnly`.
- `sort` is offered on `list-all-subscriptions` (via `SortParameter`), but the spec exposes no `sort` query param there (only the customer-scoped list supports it).

## Reference

- [Get subscription](https://docs.mollie.com/reference/get-subscription)
- [Create subscription](https://docs.mollie.com/reference/create-subscription)
- [Update subscription](https://docs.mollie.com/reference/update-subscription)
- [Cancel subscription](https://docs.mollie.com/reference/cancel-subscription)
- [List customer subscriptions](https://docs.mollie.com/reference/list-subscriptions)
- [List all subscriptions](https://docs.mollie.com/reference/list-all-subscriptions)
- [List subscription payments](https://docs.mollie.com/reference/list-subscription-payments)

## Verification

- `yarn build:declarations` (tsc over the full type graph) — passes
- `eslint` on touched dirs — clean
- `prettier` — clean

### DIFF
--- a/src/binders/customers/subscriptions/CustomerSubscriptionsBinder.ts
+++ b/src/binders/customers/subscriptions/CustomerSubscriptionsBinder.ts
@@ -33,7 +33,7 @@ export default class CustomerSubscriptionsBinder extends Binder<SubscriptionData
    * -   `amount[currency]="EUR" amount[value]="10.00" interval="1 month" startDate="2018-04-30"` Your consumer will be charged €10 on the last day of each month, starting in April 2018.
    *
    * @since 1.3.2
-   * @see https://docs.mollie.com/reference/v2/subscriptions-api/create-subscription
+   * @see https://docs.mollie.com/reference/create-subscription
    */
   public create(parameters: CreateParameters): Promise<Subscription>;
   public create(parameters: CreateParameters, callback: Callback<Subscription>): void;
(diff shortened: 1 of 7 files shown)


edorivai (approved):
Reviewed: additive/non-breaking. `SubscriptionLinks.mandate?` optional, `profileId?` optional on subscription-payments page params, and the new `getMandate()` helper faithfully copies the `getProfile` pattern (imports resolve). `SubscriptionStatus` and interval/amount/times untouched. `@since 4.6.0` matches the next minor. LGTM.

PR #512, opened 2026-06-30
feat(applePay): add profileId to payment session request

janpaepke:
## Summary

Syncs the Apple Pay wallet types against the Mollie OpenAPI spec (`request-apple-pay-payment-session`).

- Add `profileId?: string` to `RequestPaymentSessionParameters`. Required for organization-level credentials (e.g. OAuth access tokens), omitted for single-profile API keys. Modeled inline (request-only here — the response is the opaque Apple Pay session object), consistent with how the other binders expose `profileId`.
- Refresh three `@see` links from the legacy `/reference/v2/wallets-api/...` format to the current `/reference/...` format — the `validationUrl` and `domain` params plus the `requestPaymentSession` binder method.

All non-breaking: one added optional field plus doc-link updates.

## Docs reference

https://docs.mollie.com/reference/request-apple-pay-payment-session

## Verification

- `yarn build:declarations` ✓ (tsc over the full type graph)
- `eslint` ✓
- `prettier` ✓

### DIFF
--- a/src/binders/applePay/ApplePayBinder.ts
+++ b/src/binders/applePay/ApplePayBinder.ts
@@ -30,7 +30,7 @@ export default class ApplePayBinder {
    * For the full documentation, see the official [Apple Pay JS API](https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_js_api) documentation.
    *
    * @since 3.5.0
-   * @see https://docs.mollie.com/reference/v2/wallets-api/request-apple-pay-payment-session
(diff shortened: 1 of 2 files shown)


edorivai (approved):
Reviewed: `profileId?: string` added optional to `RequestPaymentSessionParameters`, and since `requestPaymentSession` POSTs the whole params object as the body, it's actually transmitted — no extra plumbing needed. Bundled `@see` URL refresh is harmless. LGTM.

